PR #67, opened 2022-01-19
Jerlich

jerlich:
I have created some helper functions for extracting data and plotting. 

There is also a notebook called `social` which uses some of these functions and a file called `arena0.py` which acts as a helper function to run some plot generation and data export jobs..


jkbhagatio:
gonna move this to 'aeon_analysis' repo today

PR #118, opened 2022-06-02
Datajoint pipeline

### LINKED ISSUE #88: DataJoint - data ingestion of the 24/7 experiment (started on 2/23/2022)
- [ ] Incorporate ingestion of per-acquisition-epoch config files (YAML) for 24/7 experiment
- [ ] Experiment meta info insertion (e.g. experiment name, experiment subjects, etc.)

### LINKED ISSUE #100: DataJoint - adapt ingestion routine to the new `aeon.io` device/stream loader mechanism
From `reorg` branch

Part of this effort is also to specify DeviceType and Device - a generic mechanism for data ingestion based on the API's load() for any new devices

iamamutt:
Incorporate DataJoint pipeline changes into main branch

Fix #88 
Fix #100


.gitignore
```diff
@@ -24,3 +24,4 @@ dj_local_conf.json
 log*.txt
 scratch/
 scratch*.py
+**/*.nfs*
```
jkbhagatio, line 27: what is this ignore for? just curious

iamamutt: Not sure why but sciviz generates these process files when it's running, e.g, `aeon_mecha/aeon/dj_pipeline/sciviz/.nfs000000000178f2a400000056`

iamamutt:
- Update pyproject.toml to remove poetry specific items
- Should be able to install the package using `pip install .` or `pip install -e .`
  - If using conda, run `mamba env create -qf env.yml` before `pip install .`

Note: 

- `blas` not available on pypi, so only added in `env.yml`
- `opencv` not available on pypi, so changed to `opencv-python` in both `pyproject.toml` and `env.yml` files. @glopesdev Can you verify that this version is okay?

env.yml
```diff
@@ -7,47 +7,41 @@ channels:
...
+  - pre-commit # ................... (dev)
+  - pytest # ....................... (dev)
+  - pytest-cov # ................... (dev)
+  - sphinx # ....................... (dev)
+  - tox # .......................... (dev)
+  - pip:
+      - datajoint>=0.13.6, <1
+      - git+https://github.com/datajoint-company/datajoint-utilities.git
+      - git+https://github.com/Technologicat/pyan.git
+      - opencv-python
```
jkbhagatio, line 47: hey @iamamutt , is there a particular reason to install open-cv via pip from pypi rather than from conda-forge?

Also I think you took out the `gh` package - I was using this for some github cli stuff I found useful

iamamutt: @jkbhagatio I added back `gh`. I thought it was something I added on my own. 

As for `opencv-python`. `pypi` doesn't have `opencv`, but `opencv-python`, and it's the official version (https://opencv.org/opencv-python-is-now-an-official-opencv-project/). Conda has `opencv`, but I couldn't find out what the build options were to assess the differences between that and the pypi version. (e.g., no ffmpeg, contrib packages, etc..).

I could put `opencv` in the `env.yml` file, and `opencv-python` in the `pyproject.toml file`, but an install of the aeon package will trigger both packages to be installed if installing from conda then pip, which will lead to issues.

PR #143, opened 2022-08-08
add plotly visualizations 

JaerongA:
1. add functions to dj_pipeline/utils/plotting.py
2. add VisitSummaryPlot to dj_pipeline/report.py


aeon/dj_pipeline/analysis/visit_analysis.py
```diff
@@ -388,6 +388,86 @@ def make(self, key):
...
             self.FoodPatch.insert(in_food_patch_times)
 
+    @classmethod
+    def _get_foraging_bouts(
+        cls,
+        visit_per_day_row,
+        wheel_dist_crit=None,
+        min_bout_duration=None,
+        using_aeon_io=False,
+    ):
```
ttngu207, line 398: Please add a short docstring to this function, explaining what it is used for.
Also, what is the `visit_per_day_row` expected to be? A pandas series/dataframe? I think that would put too much burden on the users to be able to use this method, maybe take some sort of a `visit_day_key`?

ttngu207: Where is this class method expected to be used? If it is expected to be used by another class/function elsewhere, then this should not be a hidden `_get_foraging_bouts` method - remove the leading `_` in the method name/

aeon/dj_pipeline/analysis/visit_analysis.py
```diff
@@ -388,6 +388,86 @@ def make(self, key):
...
+            )  # in seconds
+            if ts_duration < min_bout_duration:
+                continue
+
+            wheel_start, wheel_end = ts_array[0], ts_array[-1]
+            if wheel_start > wheel_end:  # skip if timestamps were misaligned
+                continue
+
+            wheel_data = acquisition.FoodPatchWheel.get_wheel_data(
+                experiment_name="exp0.2-r0",
```
ttngu207, line 458: why hardcoding `experiment_name` here? can we retrieve it from the input (e.g. if we take a `visit_day_key` as input, we can get `experiment_name` from there)

aeon/dj_pipeline/report.py
```diff
@@ -442,6 +448,68 @@ def delete_outdated_plot_entries():
         tbl.delete_outdated_entries()
 
 
+@schema
+class VisitSummaryPlot(dj.Computed):
```
ttngu207, line 452:
```suggestion
class VisitDailySummaryPlot(dj.Computed):
```

aeon/dj_pipeline/report.py
```diff
@@ -10,12 +10,18 @@
...
 
 from . import acquisition, analysis, get_schema_name
 
 schema = dj.schema(get_schema_name("report"))
 os.environ["DJ_SUPPORT_FILEPATH_MANAGEMENT"] = "TRUE"
 
+experiment_name = "exp0.2-r0"
+MIN_VISIT_DURATION = 24  # in hours (minimum duration of visit for analysis)
+WHEEL_DIST_CRIT = 1  # in cm (minimum wheel distance travelled)
+MIN_BOUT_DURATION = 1  # in seconds (minimum foraging bout duration)
```
ttngu207, line 24: Let's not make these global variables in this module, you can just define them wherever they're actually used.

aeon/dj_pipeline/report.py
```diff
@@ -442,6 +448,68 @@ def delete_outdated_plot_entries():
...
+    ---
+    pellet_count_plotly:             longblob  # Dictionary storing the plotly object (from fig.to_plotly_json())
+    wheel_distance_travelled_plotly: longblob
+    total_distance_travelled_plotly: longblob  
+    foraging_bouts_plotly:           longblob  
+    """
+
+    key_source = dj.U("experiment_name", "subject", "visit_start", "visit_end") & (
+        VisitEnd
+        & f'experiment_name="{experiment_name}"'
```
ttngu207, line 464: I don't think we need to apply a restriction on `experiment_name` here.
This is downstream from `VisitSummary` - meaning for each entry in VisitSummary, we should be able to make a daily summary plot for it.
Unless you're intended to restrict this to only multi-day visit? if so we can specify that in visit_duration

aeon/dj_pipeline/utils/plotting.py
```diff
@@ -197,3 +199,144 @@ def plot_average_time_distribution(session_keys):
     fig.update_xaxes(tickangle=45)
 
     return fig
+
+
+def plot_summary_per_visit(
```
ttngu207, line 204:
```suggestion
def plot_visit_daily_summary(
```

aeon/dj_pipeline/utils/plotting.py
```diff
@@ -197,3 +199,144 @@ def plot_average_time_distribution(session_keys):
...
+        >>> fig = plot_summary_per_visit(visit_key, attr='wheel_distance_travelled', per_food_patch=True)
+        >>> fig = plot_summary_per_visit(visit_key, attr='total_distance_travelled')
+    """
+
+    subject, visit_start = (
+        visit_key["subject"],
+        visit_key["visit_start"],
+    )
+
+    per_food_patch = False if attr.startswith("total") else True
```
ttngu207, line 230:
```suggestion
    per_food_patch = not attr.startswith("total")
```

aeon/dj_pipeline/utils/plotting.py
```diff
@@ -197,3 +199,144 @@ def plot_average_time_distribution(session_keys):
...
+    visit_per_day_df["subject"] = "_".join([subject, visit_start.strftime("%m%d")])
+    visit_per_day_df["day"] = (
+        visit_per_day_df["visit_date"] - visit_per_day_df["visit_date"].min()
+    )
+    visit_per_day_df["day"] = visit_per_day_df["day"].dt.days
+    visit_per_day_df["foraging_bouts"] = visit_per_day_df.apply(
+        VisitTimeDistribution._get_foraging_bouts,
+        args=(wheel_dist_crit, min_bout_duration, using_aeon_io),
+        axis=1,
+    )
```
ttngu207, line 321: now I see how `VisitTimeDistribution._get_foraging_bouts()` is used. If this is the case, then perhaps no need to make this a class method, just let this be a private function here in this `plotting` module, and call it here.
Because in order to use it, ones would need to have a very specifically made pandas dataframe as input.

ttngu207 (changes requested):

aeon/dj_pipeline/analysis/visit_analysis.py
JaerongA: Yeah, good point. I just designed this way to use the `df.apply` method, but it makes things harder to read or debug. Making it a standalone helper function should be better

aeon/dj_pipeline/report.py
JaerongA: these functions are called within make methods. You want these values to be hard-coded as an argument?

ttngu207: Yes, either hardcode or make them variable scoped within the `make()` only, should not be global variables.

aeon/dj_pipeline/report.py
```diff
@@ -468,24 +468,24 @@ class VisitSummaryPlot(dj.Computed):
     def make(self, key):
         from aeon.dj_pipeline.utils.plotting import (
             plot_foraging_bouts,
-            plot_summary_per_visit,
+            plot_visit_daily_summary,
         )
```
ttngu207, line 472: Should we move these imports to the beginning of the module? Are there reasons for these to be imported only within this `make()`? (avoiding circular dependency?)

aeon/dj_pipeline/report.py
JaerongA: Isn't this plotting for `'exp0.2-r0'`  only?

aeon/dj_pipeline/report.py
JaerongA: It's just cuz they are util functions to be used only within that method..  circular dependency wasn't an issue.

aeon/dj_pipeline/report.py
```diff
@@ -459,9 +455,11 @@ class VisitDailySummaryPlot(dj.Computed):
     foraging_bouts_plotly:           longblob  
     """
 
+    MIN_VISIT_DURATION = 24  # in hours (minimum duration of visit for analysis)
```
ttngu207, line 458:
```suggestion
   _min_visit_duration = 24  # in hours (minimum duration of visit for analysis)
```

aeon/dj_pipeline/report.py
```diff
@@ -459,9 +455,11 @@ class VisitDailySummaryPlot(dj.Computed):
...
+
     key_source = dj.U("experiment_name", "subject", "visit_start", "visit_end") & (
         VisitEnd
-        & f'experiment_name="{experiment_name}"'
+        & f"experiment_name= 'exp0.2-r0'"
         & f"visit_duration > {MIN_VISIT_DURATION}"
```
ttngu207, line 463:
```suggestion
        & f"visit_duration > {_min_visit_duration}"
```

ttngu207 (approved):

Issue #90, opened 2022-03-15
Mismatch in weight-scale folder and device name


ttngu207:
@glopesdev , @jkbhagatio 

Take the data in this epoch for example: /ceph/aeon/test2/experiment0.2/2022-03-01T23-36-39

The `Metadata.yml` specifies the weight-scale device to have the name `WeightNest`, while the corresponding data folder is named `Nest`. Due to this mismatch, calling the api to load weight measurement will return empty dataframe

```
api.weightdata("/ceph/aeon/test2/experiment0.2", device="WeightNest")

ttngu207:
Hi @jkbhagatio , @glopesdev , just want to follow up on this issue

jkbhagatio:
I believe this particular issue was fixed in the specified 'metadata.yml' file. Also as a fix going forward on the acquisition side, the device names in metadata.yml are now generated directly from the acquisition device names, so this type of mismatch should no longer be able to occur.

PR #148, opened 2022-08-19
🐛 fixed a bug in foodpatch device mapping in exp02

JaerongA:
Related to this : #90 

Fixes the food patch device mapping error

### DIFF
--- a/aeon/dj_pipeline/acquisition.py
+++ b/aeon/dj_pipeline/acquisition.py
@@ -1,19 +1,18 @@
-import datajoint as dj
 import datetime
 import pathlib
+
+import datajoint as dj
 import numpy as np
 import pandas as pd
 
-from aeon.io import api as io_api
-from aeon.schema import dataset as aeon_schema
-from aeon.io import reader as io_reader
 from aeon.analysis import utils as analysis_utils
+from aeon.io import api as io_api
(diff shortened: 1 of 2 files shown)


aeon/dj_pipeline/acquisition.py
```diff
@@ -955,41 +954,56 @@ def key_source(self):
...
+        raw_data_dir = Experiment.get_data_directory(key, directory_type=dir_type)
+
         weight_scale_description = (ExperimentWeightScale & key).fetch1(
             "weight_scale_description"
         )
 
-        raw_data_dir = Experiment.get_data_directory(key, directory_type=dir_type)
-
-        device = getattr(
+        weight_device = getattr(
```
ttngu207, line 968: Can we keep the previous pattern?

```
device = getattr(
_device_schema_mapping[key["experiment_name"]], weight_scale_description
        )
```

and use `device.WeightRaw` afterward. 

There may be other data than just `WeightRaw` for the device.

ttngu207: and keep the logic consistent with other `make()` functions

JaerongA: ok, I'll make those changes.

ttngu207 (approved):

PR #173, opened 2022-11-10
fix: 🐛 fix message log ingestion error

JaerongA:
1. skip ingesting ExperimentLog.Message if it errors out in the aeon api loader.
2. Don't apply maintenance filter if the log doesn't exist in the dj table, treating all as experiment.

### DIFF
--- a/aeon/dj_pipeline/acquisition.py
+++ b/aeon/dj_pipeline/acquisition.py
@@ -11,8 +11,8 @@ from aeon.io import reader as io_reader
 from aeon.schema import dataset as aeon_schema
 
 from . import get_schema_name
-from .utils.load_metadata import extract_epoch_metadata, ingest_epoch_metadata
 from .utils import paths
+from .utils.load_metadata import extract_epoch_metadata, ingest_epoch_metadata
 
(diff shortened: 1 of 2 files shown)


aeon/dj_pipeline/acquisition.py
```diff
@@ -612,54 +612,62 @@ class Message(dj.Part):
...
+                        "message_time": r.name,
+                        "message": r.message,
+                        "message_type": r.type,
+                    }
+                    for _, r in log_messages.iterrows()
+                ),
+                skip_duplicates=True,
+            )
+        except:
+            print("Can't read from device.MessageLog")
```
ttngu207, line 644: Let's catch specific error type(s) (e.g. `except ValueError, KeyError` something like that) - instead of an open ended `except`. 
Also, `logger.info()` or `logger.warning` is better than `print` if applicable

JaerongA: yes, that should be better.

ttngu207 (approved):

ttngu207 (approved):

Issue #181, opened 2022-12-12
Failure in loading Octagon OSC file


ygrayson:
Hello,

As I was trying to load the Octagon OSC files with the following lines:

`from aeon.schema.dataset import octagon01`
`root = "/ceph/aeon/aeon/data/raw/OCTAGON01/conf1/"`
`slices = api.load(root, reader=octagon01.OSC.Slice, epoch='2022-10-28T15-22-42')`

The following error poped up:
<img width="873" alt="image" src="https://user-images.githubusercontent.com/33047668/207092914-b03e4aa3-e351-4c87-8755-65b02a1eb93d.png">

This does not present any problem while loading `epoch='2022-08-05T08-54-05'`. I checked the CSV file and it seems like the later epoch has one more colomn than the earlier epoch, which I think could be the cause of this bug.

Thanks.

jkbhagatio:
Closing this as was fixed with a column reordering on writing out the csv file

PR #183, opened 2022-12-22
debug OSC.Slice reading csv file

ygrayson:
As described in issue #181, the Octagon OSC messages have changed sometime in August and there is one extra column in the OSC_octagonslice csv file and one extra column in the OSC_octagongratingsslice csv file. The API need to be updated to correctly read in these files. This merge request fix this issue.

However, this fix does not allow backward compatibility. Files after August will be read correctly but OSC files before that will not. Please feel free to review the issue and accept/reject the change.

Thanks.

### DIFF
--- a/aeon/schema/octagon.py
+++ b/aeon/schema/octagon.py
@@ -24,7 +24,7 @@ class OSC:
             'typetag',
             'wall_id',
             'r', 'g', 'b', 'a',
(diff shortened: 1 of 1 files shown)


ygrayson:
I have also added 2 methods for the new OSC files in the second commit.

PR #193, opened 2023-03-29
fix bugs

JaerongA:
- [x] Fix docker image url.
- [x] Roll back sciviz to fix entry form rendering error.

### DIFF
--- a/aeon/dj_pipeline/webapps/sciviz/docker-compose-local.yaml
+++ b/aeon/dj_pipeline/webapps/sciviz/docker-compose-local.yaml
@@ -7,7 +7,7 @@ services:
   pharus:
     cpus: 2.0
     mem_limit: 4g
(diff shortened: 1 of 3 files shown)


aeon/dj_pipeline/streams.py
```diff
@@ -15,17 +15,15 @@
 
 logger = dj.logger
 
-
-schema_name = f'u_{dj.config["database.user"]}_streams'  # for testing
-# schema_name = get_schema_name("streams")
+schema_name = get_schema_name("streams")
```
ttngu207, line 18: Is this fully tested and ready for production deployment?

JaerongA: not with `presocial` yet..  but there tables being created under `aeon_test_streams` by the worker.

PR #195, opened 2023-04-03
Ensure each chunk is compared against filter bounds

### LINKED ISSUE #186: Metadata.yml and other unchunked files cannot be filtered by time range
As ``chunk_key`` here returns the epoch timestamp directly
https://github.com/SainsburyWellcomeCentre/aeon_mecha/blob/bf96b6346f272dc3a7d91ea636e5f93634dbeeae/aeon/io/api.py#L37-L46

and the ``chunk_range`` only includes hourly timestamps, files that do not have hourly timestamps  in their names (e.g. Metadata.yml) are removed by the filter.
https://github.com/SainsburyWellcomeCentre/aeon_mecha/blob/bf96b6346f272dc3a7d91ea636e5f93634dbeeae/aeon/io/api.py#L118-L119

MWE:
```python
import aeon.io.api as aeon
from aeon.schema.dataset import exp02

root = "/ceph/aeon/aeon/data/raw/AEON3/presocial0.1/"
aeon.load(root, exp02.Metadata, start=pd.Timestamp("2023-02-24 00:00:00"), end=pd.Timestamp("2023-02-26 00:00:00"))
```

### LINKED ISSUE #194: Load does not accept partial time range filters
The low-level reader API does not currently accept a partially specified time range filter. For example:

```python
aeon.load(root, reader, start=start, end=None)
```

glopesdev:
This PR refactors the approach to filtering chunks. Instead of generating a chunk range and comparing each item in the list against each chunk name, we now simply go through the list of all chunks and compare each against the chunked filter boundaries.

This allows correct filtering of non-chunked data and also the specification of partial time filters.

Fixes #186 
Fixes #194


tests/io/test_api.py
```diff
@@ -0,0 +1,42 @@
...
+
+@mark.api
+def test_load_monotonic():
+    data = aeon.load('./tests/data/monotonic', exp02.Patch2.Encoder)
+    assert data.index.is_monotonic_increasing
+
+@mark.api
+def test_load_nonmonotonic():
+    data = aeon.load('./tests/data/nonmonotonic', exp02.Patch2.Encoder)
+    assert not data.index.is_monotonic_increasing
```
lochhh, line 39:
```suggestion
@pytest.mark.parametrize(
    "args,kwargs,expected",
    [
        pytest.param(
            ["./tests/data/nonmonotonic", exp02.Patch2.Encoder],
            {"start": pd.Timestamp("2022-06-06T13:00:49")},
            0,
            id="start_only",
            marks=mark.api,
        ),
        pytest.param(
            ["./tests/data/nonmonotonic", exp02.Patch2.Encoder],
            {"end": pd.Timestamp("2022-06-06T13:00:49")},
            0,
            id="end_only",
            marks=mark.api,
        ),
        pytest.param(
            ["./tests/data/nonmonotonic", exp02.Metadata],
            {"start": pd.Timestamp("2022-06-06T09:00:00")},
            0,
            id="non_chunked",
            marks=mark.api,
        ),
        pytest.param(
            ["./tests/data/monotonic", exp02.Patch2.Encoder],
            {},
            True,
            id="monotonic",
            marks=mark.api,
        ),
        pytest.param(
            ["./tests/data/nonmonotonic", exp02.Patch2.Encoder],
            {},
            False,
            id="nonmonotonic",
            marks=mark.api,
        ),
    ],
)
def test_load(args, kwargs, expected):
    data = aeon.load(*args, **kwargs)
    if type(expected) == bool:
        assert data.index.is_monotonic_increasing == expected
    else:
        assert len(data) > expected
```

Since these tests are testing the same ``load`` function with various inputs, we could also consider [parameterised tests](https://docs.pytest.org/en/latest/how-to/parametrize.html#parametrize-basics).

glopesdev: I feel parameterised tests work best when each set of parameters fits in a line so you can massively reduce duplication. In this case however since there are so many different parameters to vary it looks like the parameterised version takes up even more lines than the original code.

PR #205, opened 2023-06-14
Fixes and improvements to the low-level API

### LINKED ISSUE #185: In `api.load()` add support for Pathlib Paths
In general for paths, we should support Path objects over strs

### LINKED ISSUE #203: Add support for reading heartbeat notifications
Now that we have access to the logs from the main clock synchronizer and all heartbeat events are turned on, inspection of the heartbeat notifications will likely play an increasing role in data QC. It should likely be added to the general list of stream readers so it can easily be included in any Harp device meta-schema.

### LINKED ISSUE #204: Harp reader throws if no list of column names is specified
This is a regression from previous versions which seems to have been introduced by the backwards compatibility layer where columns missing from the data will be padded with NaN values. Calling the `len` operator on a `None` value will throw:

https://github.com/SainsburyWellcomeCentre/aeon_mecha/blob/d52a44f2275fadb29bd024b56e00a837b9c2669f/aeon/io/reader.py#L67-L69

glopesdev:
This PR fixes a number of issues and adds a few features and readers missing from the low-level API for the purposes of data QC and possibly ingestion.

Fixes #204 
Fixes #203 
Fixes #185

### DIFF
--- a/aeon/io/api.py
+++ b/aeon/io/api.py
@@ -2,6 +2,7 @@ import bisect
 import datetime
 import pandas as pd
 from pathlib import Path
+from os import PathLike
 
 """The duration of each acquisition chunk, in whole hours."""
 CHUNK_DURATION = 1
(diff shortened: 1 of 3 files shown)


jkbhagatio:
cool, looks good to me, but let's review this in data arch meeting

jkbhagatio (approved):

PR #206, opened 2023-06-14
Add pose reader classes supporting ingestion of SLEAP tracking data

jkbhagatio:
A draft of the reader for

https://github.com/SainsburyWellcomeCentre/aeon_experiments/issues/241


aeon/schema/social.py
```diff
@@ -0,0 +1,27 @@
...
+
+
+class Sleap(_reader.Harp):
+    def __init__(self, model_path):
+        """Initializes a reader for Harp-binarized Sleap tracking data given a Sleap network model path."""
+        columns = ["class", "class_likelihood"]
+        # Get bodyparts from Sleap config.
+        with open(model_path) as f:
+            config = json.load(f)
+        bodyparts = config["model"]["heads"]["multi_class_topdown"]["confmaps"]["bodyparts"]
```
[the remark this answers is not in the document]
jkbhagatio: @lochhh also suggested using "data", "labels", "skeletons"..."py/tuple" for the json key to get the body parts

ttngu207:
One potential change to the Sleaper reader class is to remove the requirement of `model_path` as an argument to instantiate the class. 
Instead, this `model_path` can be specified upload loading the data. 

I will review and potentially propose a modification to the current solution.

ttngu207:
We discussed the possibility of the reader not needing the model path. The model path can be extract from some meta information somewhere.
This model path information is stored anyway, perhaps better yet that we store a timestamp and the model path as event, for each time the model is changed.

jkbhagatio:
Given the above, we assume that model metadata config file lives next to the model? 

@glopesdev also if we go with the dataframe columns being:

class | class_conf | part | part_conf | x | y
-------------------------------------------- 

will need to slightly change this reader to output in this format

jkbhagatio:
Hi all, there should be a working version of this now, via the `Pose` reader.

It should work with the example files in '/ceph/aeon/aeon/data/processed/test-enc1...'

We should add a proper test for this (and other readers), and decide where it will finally live (see https://github.com/SainsburyWellcomeCentre/aeon_mecha/issues/226), but this provides a working solution for now.

aeon/schema/social.py
```diff
@@ -0,0 +1,94 @@
...
+    def __init__(self, pattern):
+        self.pattern = pattern
+        self.extension = "bin"
+
+    def read(self, file):
+        """
+        Reads data from the Harp-binarized tracking file
+        """
+        # Get config file from `file`, then bodyparts from config file.
+        ceph_proc_dir = "/ceph/aeon/aeon/data/processed"
```
glopesdev, line 22: We will need a more general way to search for the root processed folder before we can move this reader to the `io` core module. Most likely this will come from resolution of #227 but I agree that for now this would provide a workable specific solution for the coming experiments.

aeon/schema/social.py
```diff
@@ -0,0 +1,94 @@
...
+        model_dir = file.stem.replace("_", "/")
+        config_file_dir = Path(ceph_proc_dir + "/" + model_dir)
+        assert config_file_dir.exists(), f"Cannot find model dir {config_file_dir}"
+        config_file = get_config_file(config_file_dir)
+        parts = self.get_bodyparts(config_file)
+        if parts is None:
+            raise ValueError(f"Cannot find bodyparts in {config_file}.")
+        # With bodyparts, set columns as set by Bonsai, and read data in default format.
```
glopesdev, line 30: Minor meta-comment, but I would generalize this comment to the Harp format, e.g.:

`# Using bodyparts, assign column names to the Harp raw register values, and read data in default format.`

aeon/schema/social.py
```diff
@@ -0,0 +1,94 @@
...
+        """
+        Reads data from the Harp-binarized tracking file
+        """
+        # Get config file from `file`, then bodyparts from config file.
+        ceph_proc_dir = "/ceph/aeon/aeon/data/processed"
+        model_dir = file.stem.replace("_", "/")
+        config_file_dir = Path(ceph_proc_dir + "/" + model_dir)
+        assert config_file_dir.exists(), f"Cannot find model dir {config_file_dir}"
+        config_file = get_config_file(config_file_dir)
+        parts = self.get_bodyparts(config_file)
```
glopesdev, line 27: It might be useful to cache the config file bodyparts in the reader itself to avoid having to reload and parse the same config file across multiple chunks in a long session. One consideration here is when should the cache be invalidated. There is no good infrastructure design to support this currently, but a natural decision would be restricting the cache to a single call to "load".

aeon/schema/social.py
```diff
@@ -0,0 +1,94 @@
...
+        self.columns = columns
+        data = super().read(file)
+        # Set new columns, and reformat `data`.
+        columns = ["class", "class_likelihood", "part", "part_likelihood", "x", "y"]
+        new_data = pd.DataFrame(columns=columns)
+        for part in parts:
+            part_data = data[["class", "class_likelihood", f"{part}_likelihood", f"{part}_x", f"{part}_y"]]
+            part_data.insert(2, "part", part)
+            part_data.columns = columns
+            new_data = pd.concat([new_data, part_data])
```
glopesdev, line 43: It might be more memory and compute performant here to simply keep a list with all the part data tables and then do a single call to `pd.concat` to avoid allocating intermediate data frames.

aeon/schema/social.py
```diff
@@ -0,0 +1,94 @@
...
+        """Returns a list of bodyparts from a model's config file."""
+        parts = None
+        if file.stem == "confmap_config":  # SLEAP
+            with open(file) as f:
+                config = json.load(f)
+            try:
+                heads = config["model"]["heads"]
+                for model in heads.keys():
+                    if heads[model] is not None:
+                        parts = heads[model]["confmaps"]["part_names"]
```
glopesdev, line 56: If I remember correctly from development of the SLEAP package, the config file structure might change depending on the model type used (i.e. top-down centroid, top-down with identity, etc) so this might require in the future more flexibility to take into account which model type we are using.

aeon/schema/social.py
```diff
@@ -0,0 +1,94 @@
...
+            part_data = data[["class", "class_likelihood", f"{part}_likelihood", f"{part}_x", f"{part}_y"]]
+            part_data.insert(2, "part", part)
+            part_data.columns = columns
+            new_data = pd.concat([new_data, part_data])
+        return new_data.sort_index()
+       
+    def get_bodyparts(file):
+        """Returns a list of bodyparts from a model's config file."""
+        parts = None
+        if file.stem == "confmap_config":  # SLEAP
```
glopesdev, line 49: Would be nice to reuse the string in `config_file_names` here so that we don't risk in the future having typos when comparing config file names for different models.

aeon/schema/social.py
```diff
@@ -0,0 +1,94 @@
...
+                        parts = heads[model]["confmaps"]["part_names"]
+                        break
+            except KeyError as err:
+               raise KeyError(f"Cannot find bodyparts in {file}.") from err
+        return parts
+
+
+def get_config_file(
+    config_file_dir,
+    config_file_names=[
```
glopesdev, line 65: We might want to have this be in the outer scope so we can query the map in other bits of code.

aeon/schema/social.py
```diff
@@ -0,0 +1,94 @@
...
+        if (config_file_dir / f).exists():
+            config_file = config_file_dir / f
+            break
+    assert config_file is not None, f"Cannot find config file in {config_file_dir}"
+    return config_file
+
+def class_int2str(tracking_df, config_file_dir):
+    """Converts a class integer in a tracking data dataframe to its associated string (subject id)."""
+    config_file = get_config_file(config_file_dir)
+    if config_file.stem == "confmap_config":  # SLEAP
```
glopesdev, line 81: Same as above, reuse `config_file_names` map.

glopesdev (commented):
Looks great, just have a few comments to look into before merging. I think most of them can be left for when we decide to move this reader into the core `io` module, but some could be nice to do right now.

aeon/schema/social.py
jkbhagatio: yep sounds good, will resolve this for now then

aeon/schema/social.py
jkbhagatio: sounds good, will do this in the next commit

aeon/schema/social.py
jkbhagatio: hmm, I think if we do as you suggest, we'd need to index into `config_file_names` here, so we'd look for a match via an index via a string match. I kind of prefer looking for a string match, but let's discuss.

aeon/schema/social.py
```diff
@@ -0,0 +1,94 @@
...
+            with open(file) as f:
+                config = json.load(f)
+            try:
+                heads = config["model"]["heads"]
+                for model in heads.keys():
+                    if heads[model] is not None:
+                        parts = heads[model]["confmaps"]["part_names"]
+                        break
+            except KeyError as err:
+               raise KeyError(f"Cannot find bodyparts in {file}.") from err
```
lochhh, line 59: Can confirm that "part_names" may exist on either level 2 or level 3 in different models:

- `heads[model]["part_names"]` for "single_instance", "centered_instance"
- `heads[model]["confmaps"]["part_names"]` for models with prefix "multi_"
- not at all for "centroid"
```suggestion
            try:
                    heads = config["model"]["heads"]
                    for model, conf in heads.items():
                        if conf is not None:
                            if "part_names" in conf.keys():
                                parts = conf["part_names"]
                            else:
                                parts = heads[model]["confmaps"]["part_names"]
            except KeyError as err:
               raise KeyError(f"Cannot find bodyparts in {file}.") from err
```

aeon/schema/social.py
```diff
@@ -0,0 +1,88 @@
...
+        data = super().read(file)
+        # Set new columns, and reformat `data`.
+        new_columns = ["class", "class_likelihood", "part", "part_likelihood", "x", "y"]
+        new_data = pd.DataFrame(columns=new_columns)
+        part_data_list = [None] * len(parts)
+        for i, part in enumerate(parts):
+            part_columns = ["class", "class_likelihood", f"{part}_x", f"{part}_y", f"{part}_likelihood"]
+            part_data_list[i] = data[part_columns]
+            part_data_list[i].insert(2, "part", part)
+            part_data_list[i].columns = new_columns
```
glopesdev, line 41: Here I would probably just take a temp variable for `data[part_columns]` and then manipulate that object directly before using `append` to add it to the list.

aeon/schema/social.py
```diff
@@ -0,0 +1,88 @@
...
+        new_data = pd.concat(part_data_list)
+        return new_data.sort_index()
```
glopesdev, line 43: Why is the sort necessary? Can't we assume that parts are inserted in the other they arrive in the `bin` file? It shouldn't really matter anyway.

aeon/schema/social.py
```diff
@@ -0,0 +1,88 @@
...
+def get_config_file(
+    config_file_dir,
+    config_file_names=[
+        "confmap_config.json",  # SLEAP (add others for other trackers to this list)
+    ],
+):
+    """Returns the config file from a model's config directory."""
+    config_file = None
+    for f in config_file_names:
+        if (config_file_dir / f).exists():
```
glopesdev, line 68: Does this really do what I think it is doing `/` does it really append a slash to the path? Crazy stuff...

glopesdev (changes requested):
A few more optimization and formatting questions, but otherwise looks good!

aeon/schema/social.py
jkbhagatio: The way parts are added, you add all the timestamps for one part, then all the timestamps for next part, etc. 

So the frame is sorted by parts, not by time, without this `sort_index()`

aeon/schema/social.py
```diff
@@ -0,0 +1,88 @@
...
+        self.pattern = pattern
+        self.extension = "bin"
+
+    def read(self, file, ceph_proc_dir="/ceph/aeon/aeon/data/processed"):
+        """Reads data from the Harp-binarized tracking file."""
+        # Get config file from `file`, then bodyparts from config file.
+        model_dir = file.stem.replace("_", "/")
+        config_file_dir = Path(ceph_proc_dir + "/" + model_dir)
+        assert config_file_dir.exists(), f"Cannot find model dir {config_file_dir}"
+        config_file = get_config_file(config_file_dir)
```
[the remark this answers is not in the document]
jkbhagatio: Yeah I've done both ways, now I generally prefer this way, with blank lines adhering only to what's explicitly suggested by pep8 specs (between functions, imports, etc.)

Now in longer codeblocks I create sections and subsection in html style syntax with `<s`, and close with `/s>`. You may have noticed this previously.

I looked to see what google style docs recs: https://google.github.io/styleguide/pyguide.html#s3.4-indentation
but they've taken it out ahah, it should be section 3.5 Maybe they didn't want the extra whitespace =p

Can revisit this in the next data architecture meeting I guess and see what people prefer - happy to make a ruling on blank lines and stick with it one way or the other.

aeon/schema/social.py
```diff
@@ -0,0 +1,88 @@
...
+        # Using bodyparts, assign column names to Harp register values, and read data in default format.
+        columns = ["class", "class_likelihood"]
+        for part in parts:
+            columns.extend([f"{part}_x", f"{part}_y", f"{part}_likelihood"])
+        self.columns = columns
+        data = super().read(file)
+        # Set new columns, and reformat `data`.
+        new_columns = ["class", "class_likelihood", "part", "part_likelihood", "x", "y"]
+        new_data = pd.DataFrame(columns=new_columns)
+        part_data_list = [None] * len(parts)
```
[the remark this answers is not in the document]
jkbhagatio: It is actually quicker to preallocate it this way, though essentially negligible in this case. "Avoid list appends like the plague" is generally a good rule to follow for code performance in python.

aeon/schema/social.py
jkbhagatio: again, can see my above comment about performancy of list appends in python.

aeon/schema/social.py
jkbhagatio: aha yup, `/` is overloaded for Path objects to join paths.

aeon/util.py
```diff
@@ -0,0 +1,20 @@
...
+
+from typing import Union
+
+
+def find_nested_key(obj: Union[dict, list], key: str):
+    """Returns the value of the first found nested key."""
+    if isinstance(obj, dict):
+        for k, v in obj.items():
+            if k == key:  # Found it!
+                return v
```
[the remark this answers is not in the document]
jkbhagatio: oho, I did not know this about `get` for dicts. Then `[ ]` are not simply calling the getter, what the heck? Weird they decided to make `[ ]` throw KeyErrors and not the getter.

But anyway cool good to know, will change this!

aeon/schema/social.py
```diff
@@ -14,12 +14,11 @@ def __init__(self, pattern):
         self.pattern = pattern
         self.extension = "bin"
 
-    def read(self, file):
+    def read(self, file, ceph_proc_dir="/ceph/aeon/aeon/data/processed"):
```
jkbhagatio, line 19: The name of `ceph_proc_dir` should be changed to something like maybe `root_to_model_dir`

aeon/schema/social.py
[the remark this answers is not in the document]
glopesdev: I see, I guess this is because you cannot initialise a list with a starting capacity as you can in C#.

In that case I wonder if writing this as a list comprehension might be even faster? When I have this kind of code where lists are generated from other lists I tend to prefer comprehensions in general, sometimes defining aux functions if needed for clarity.

That should keep readability without compromising on performance. Maybe it is just my aesthetic bias that it doesn't feel right to me to initialise and manipulate list insertions like this.

aeon/schema/social.py
glopesdev: Considering the above I would probably have written the whole concatenate with a comprehension.

aeon/schema/social.py
glopesdev: I'm still not sure about this default parameter, I think this should be solved in coordination with the issue on having a context root object. In one possible design the read method might evolve to have a default root=None or context=None parameter from which you would pull other properties, rather than having custom parameters like this.

Given this is not decided I would err on the side of not forcing a signature on read yet and just have it as a hardcoded path (or a module global if you really want to allow changing it) at this point.

aeon/schema/social.py
jkbhagatio: I've seen this pattern of initialising lists in many places, so does seem to be a common Python pattern.

Can you provide an example of how you'd do this with a comprehension?

I think if you want to use a comprehension, you'd want to include the insertion of the `"part"` column and the assignment of `new_columns` in the comprehension, and for this I think you'd need a lambda, and this wouldn't be more performant and for me I don't think it would be more readable. But maybe you're thinking of another way to write this?

aeon/schema/social.py
jkbhagatio: I think I'd like to keep this as a default arg for now. As a default arg, it won't impact `api.load()`, and it allows for easier testing (just set the arg to a test data path), which you can't do if it's hardcoded in the function body 

We could potentially get around this with a module global, but in addition to global modules generally being frowned upon, then we would also need to set the global in the test module, and so if we make future changes like you suggest we'd need to change something in two places, rather than just removing the default arg, right?

aeon/schema/social.py
glopesdev: I see, that makes sense but I still believe there must be a more efficient way to write this without the need for sorting.

The reason is that if you look at a sequence of parts in a harp message, it will have the shape: [p1t0x, p1t0y, p2t0x, p2t0y; p1t1x, p1t1y, p2t1x, p2t1y, ...]

If this was numpy, it would be possible to use reshape directly to get instead:

[p1t0x,p1t0y;
p2t0x,p2t0y;
p1t1x,p1t1y;
p2t1x,p1t1y]

This would avoid the need to sort at all. We might need to reshape before converting to pandas depending on how internally they represent columns. There are tricky bits with the repeating part names etc so needs some thought but I think it will be worth it.

jkbhagatio: I actually think there may not be a better way to do this, under the assumption we're operating from the Super's (Harp's') `read` method. One obvious naive way to do it without sorting would be to iterate through the rows in `data` and assign them one-by-one to rows in `new_data`, but ofc this is more inefficient.

I agree sorting could be avoided, but I think it may be easiest if this was done on the Bonsai/Harp side, such that harp messagaes from Sleap inference take the final form we want here `["class", "class_likelihood", "part", "part_likelihood", "x", "y"]` instead of the current form `["class", "class_likelihood", "part_x", "part_y", "part_likelihood"]`. For this I think we'd just need to add a `part` int id to the message - can we do this @glopesdev ?

jkbhagatio: Actually just thought of one other thing I could try:

generate timestamps as indices, use this to index into `data`, assign to `new_data` for each part and `n_parts` - 1 rows.

I think it'll make sense when I show the code. Will try this and show the code and profile and lyk

glopesdev: Sounds good, lmk how it goes. If necessary we can also easily split just the raw numpy loading part of the super() into a separate method, so that you would have more flexibility of manipulation before everything gets turned into a dataframe.

aeon/schema/social.py
glopesdev: Sure, sounds good it won't be much of a problem anyway. I totally agree with you on avoiding module globals, my only reasoning behind it is that currently we don't have a way of passing custom arguments into read from the load() function so if you want to use a custom root path you have to use the loader entirely from scratch.

aeon/schema/social.py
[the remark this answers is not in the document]
jkbhagatio: ok, I was overruled on this =p

aeon/schema/social.py
jkbhagatio: Ok, set this temp var

aeon/util.py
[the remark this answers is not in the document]
jkbhagatio: done

aeon/schema/social.py
jkbhagatio: @glopesdev 

This is another way I thought to do it without sorting, but it's actually ~50% longer than using the current `sort` procedure, even on large dataframes (I simulated 1 hour worth of frames at 50 fps)

```
n_parts = len(parts)
new_columns = ["class", "class_likelihood", "part", "part_likelihood", "x", "y"]
new_data = pd.DataFrame(columns=new_columns, index=np.repeat(data.index, n_parts))
for i, part in enumerate(parts):
    part_columns = ["class", "class_likelihood", f"{part}_x", f"{part}_y", f"{part}_likelihood"]
    part_data = data[part_columns]
    part_data.insert(2, "part", part)
    part_data.columns = new_columns
    new_data.iloc[i::n_parts] = part_data
```

I guess the indexing in `new_data.iloc[i::n_parts]` in python is much slower than the compiled C of `sort_index`

aeon/schema/social.py
```diff
@@ -0,0 +1,93 @@
...
+from aeon import util
+import aeon.io.reader as _reader
+
+
+class Pose(_reader.Harp):
+    """Reader for Harp-binarized tracking data given a model that outputs id, parts, and likelihoods."""
+
+    pattern: str
+    columns: List[str] = []
+    extension: str = "bin"
```
glopesdev, line 18: Do we need to redeclare class fields here instead of calling the base class constructor?

Also, initializing columns to an empty list using this syntax would make the default list shared and mutable across all class instances. I recently came across this issue in the context of data classes: https://peps.python.org/pep-0557/#mutable-default-values

In regular classes this is allowed, but probably not recommended. I realize it wouldn't be a problem the way you are using it now, but might lead to subtle and hard to track bugs down the line.

Calling the base class constructor might also be preferable here so we can use the constructor function scope to guarantee a unique empty list is assigned on creation.

In general I feel this style of declaring and initializing fields is best for data classes where you can use default factories that will be used in the auto-generated constructor to initialize a true default.

I'm happy to consider redefining the architecture of readers using data classes if needed, but in that case I would rather do it for all classes rather than opening exceptions, for consistency of reading (and opening a new issue / discussion).

aeon/schema/social.py
glopesdev: I would still like to take a jab at improving this implementation without sorting, but I'm happy to merge this as a first draft and I can open an issue / branch later for improving this.

tests/io/test_api.py
glopesdev, line 1: Can we keep this refactoring of the tests outside of this PR? I am happy discussing and refactoring the standard imports as we discussed on #238 but would make it easier to keep track of changes if that was done in a separate PR.

glopesdev (changes requested):
Looks good, just left some final comments for consideration and I think the refactoring of tests should be left to another PR, since they seem to be unrelated to testing this reader and possibly more related to the #228 PR.

jkbhagatio:
> Looks good, just left some final comments for consideration and I think the refactoring of tests should be left to another PR, since they seem to be unrelated to testing this reader and possibly more related to the #228 PR.

@glopesdev Sounds good, read all of your most recent replies/suggestions here and agree with them all - will do them all before merging, and move the tests to #228

aeon/schema/social.py
jkbhagatio: > Also, initializing columns to an empty list using this syntax would make the default list shared and mutable across all class instances.

Thanks for pointing this out. Agree with all this and made this change.

Issue #238, opened 2023-08-28
Standardize 'aeon' imports


jkbhagatio:
Different people are using different imports for common aeon modules - these imports should be standardized for readability, similar to how people typically use standard imports for other packages (e.g. `import pandas as pd` and `import numpy as np`).

One thing that should be avoided is shadow imports: e.g. I've seen the pattern `import aeon.io.api as aeon` which is bad because it shadows the entire 'aeon' package.

glopesdev:
I think this happened due to an organic evolution from the humble origins of the package when there was only a single-file `aeon` module. The fractioning came from splitting features into multiple sub-modules and folders.

This makes sense from a development perspective, but not entirely sure we should use it directly for high-level usage. For example, I would like to preserve the syntax `aeon.load` as standard, similar to pandas and numpy `np.genfromtxt` or `pd.read_csv`. I'm not completely familiar with how numpy and pandas do it, but suspect this would entail using `__init__.py` files to bring multiple sub-modules under a single namespace.

Currently as it is the `aeon` namespace itself is totally empty, and I think it might make it easier to step into the library if that module / namespace was the natural entry point, rather than having to immediately choose from a complex sub-module hierarchy.

jkbhagatio:
>  I would like to preserve the syntax aeon.load as standard, similar to pandas and numpy np.genfromtxt or pd.read_csv

Sure, yup we can do this, yes it involves setting the '__init__.py' files properly. In this case, the call function would look like this:
```
import aeon
data = aeon.load(...)
```

This is a fine solution as it avoids the shadowing problem, and we can make it a standard and discuss what other functions we may want accessible in this way.

PR #242, opened 2023-08-30
Refactored tests to use absolute file paths

jkbhagatio:
### Summary by CodeRabbit

- Refactor: Enhanced the 'aeon' top-level namespace by making the `load` function directly accessible, improving code readability and ease of use.
- Test: Updated tests in 'test_api.py' to utilize absolute file paths instead of relative ones, increasing test reliability across different environments. 
- Chore: Imported the `Path` class from the `pathlib` module for handling filesystem paths in a manner consistent with different operating systems.

### DIFF
--- a/aeon/__init__.py
+++ b/aeon/__init__.py
@@ -8,3 +8,6 @@ except PackageNotFoundError:
     __version__ = "unknown"
 finally:
     del version, PackageNotFoundError
+
+# Set functions avaialable directly under the 'aeon' top-level namespace
+from aeon.io.api import load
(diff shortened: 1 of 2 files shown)


glopesdev (commented):
This sounds good, although I would wait until we resolve #238 so we can have our beloved `aeon.load` back?

jkbhagatio:
> This sounds good, although I would wait until we resolve #238 so we can have our beloved `aeon.load` back?

@glopesdev Done - check now?

PR #271, opened 2023-11-01
Added `lab_id` and `available` to pyrat subjects

ttngu207:
- pull from pyrat: `labid`
- check and update "availability" for each animal periodically
- deprecate the use of "colony.csv"
- added table `subject.ExperimentSubject` for the manually specification of subject and experiment association (this will be a form on SciViz)

### DIFF
--- a/aeon/dj_pipeline/subject.py
+++ b/aeon/dj_pipeline/subject.py
@@ -50,7 +50,9 @@ class SubjectDetail(dj.Imported):
     -> lab.User.proj(responsible_user="user")
     -> [nullable] GeneticBackground
     -> Strain
-    cage_number: varchar(32)
+    cage_number='': varchar(32)
+    lab_id='': varchar(128)  # pyrat 'labid'
+    available=1: bool  # is this animal available on pyrat
     """
 
     def make(self, key):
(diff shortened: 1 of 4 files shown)


ttngu207:
@glopesdev , once this PR is merged, the query you want will be:
```
subject.Subject * subject.SubjectDetail & 'available = 1'
```

In SQL:
```
'SELECT `subject`,`sex`,`subject_birth_date`,`subject_description`,`lab_id`,`responsible_fullname`,`gen_bg_id`,`strain_id`,`cage_number`,`available` FROM `aeon_subject`.`subject` NATURAL JOIN `aeon_subject`.`_subject_detail` WHERE ( (available = 1))'
```

aeon/dj_pipeline/subject.py
```diff
@@ -64,27 +66,45 @@ def make(self, key):
             "eartag": eartag_or_id,
         }
         animal_resp = get_pyrat_data(endpoint=f"animals", params=params)
-        assert len(animal_resp) == 1, f"Found {len(animal_resp)} with eartag {eartag_or_id}, expect one"
-        animal_resp = animal_resp[0]
+        if not animal_resp:
```
JaerongA, line 69:
```suggestion
        if len(animal_resp) == 0:
```

I think this more explicitly filters "no animal" conditions since response may return an error dictionary.

aeon/dj_pipeline/subject.py
```diff
@@ -142,6 +162,14 @@ class SubjectComment(dj.Imported):
     """
 
 
+@schema
+class ExperimentSubject(dj.Manual):
+    definition = """
+    -> Subject
+    experiment_name: varchar(32)  # e.g. social-AEON3
```
JaerongA, line 169: would having this in a separate table make it less error-prone?

ttngu207: Not clear what you mean, what would be your suggested alternative?

JaerongA: I'm thinking about a senario of having a upstream lookup table named e.g., `Experiment` where a user can manually enter an experiment name when it's not present, and this table makes a foreign key reference to that one so that a user can just select an experiment via dropdown in SciViz, if that makes sense.

PR #310, opened 2024-01-28
Social schema

jkbhagatio:


aeon/io/device.py
```diff
@@ -1,41 +1,40 @@
...
-    If a device contains a single stream with the same pattern as the device
-    `name`, it will be considered a singleton, and the stream reader will be
-    paired directly with the device without nesting.
+    If a device contains a single stream reader with the same pattern as the device `name`, it will be 
+    considered a singleton, and the stream reader will be paired directly with the device without nesting.
 
     Attributes:
         name (str): Name of the device.
-        args (Any): Data streams collected from the device.
+        args (any): A binder function or class that returns a dictionary of Readers.
```
glopesdev, line 26: `Any` here meant the [Any type](https://docs.python.org/3/library/typing.html#the-any-type) so it should either be left with its original casing, or replaced with something else from the typing module.

A possible alternative to "binder function" could be "stream selector" or "stream accessor", since basically what the function brings in addition to the reader is the pattern to find for stream data files.

Another option is simply to call this concept a "stream". We can discuss better on a future meeting.

aeon/io/reader.py
```diff
@@ -212,7 +212,7 @@ def read(self, file):
         specified unique identifier.
         """
         data = super().read(file)
-        data = data[data.event & self.value > 0]
+        data = data[(data.event & self.value) == self.value]
```
glopesdev, line 215: I don't remember anymore what this change was about, but I feel we should leave any functional changes outside of this PR, which will already be confusing enough with all the major breaking refactoring and reorganisation.

aeon/schema/core.py
```diff
@@ -37,7 +37,7 @@ def subject_state(pattern):
     return {"SubjectState": _reader.Subject(f"{pattern}_SubjectState_*")}
 
 
-def messageLog(pattern):
+def message_log(pattern):
```
glopesdev, line 40: Can we focus this PR only on the major naming changes and terminology? I know we disagree on best practices for casing in python, but if we could leave this out of this PR so we can focus on the core discussions on naming, that would help a lot.

aeon/schema/social.py
```diff
@@ -32,7 +32,7 @@ def read(
     ) -> pd.DataFrame:
         """Reads data from the Harp-binarized tracking file."""
         # Get config file from `file`, then bodyparts from config file.
-        model_dir = Path(file.stem.replace("_", "/")).parent
+        model_dir = Path(*Path(file.stem.replace("_", "/")).parent.parts[1:])
```
glopesdev, line 35: As above, I would remove all functional changes from this PR.

glopesdev (changes requested):

aeon/io/device.py
```diff
@@ -1,41 +1,40 @@
...
 
 class Device:
-    """Groups multiple data streams into a logical device.
+    """Groups multiple Readers into a logical device.
 
-    If a device contains a single stream with the same pattern as the device
-    `name`, it will be considered a singleton, and the stream reader will be
-    paired directly with the device without nesting.
+    If a device contains a single stream reader with the same pattern as the device `name`, it will be 
+    considered a singleton, and the stream reader will be paired directly with the device without nesting.
```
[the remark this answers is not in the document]
lochhh: looks like auto-formatting

jkbhagatio:
Closing as the schema has been incorporated in #347

PR #376, opened 2024-07-03
Ensure all aeon submodules included in wheel

### LINKED ISSUE #375: Installing from main branch using pip reports missing packages
### To reproduce

On a clean Python 3.12 install run:
```
pip install git+https://github.com/SainsburyWellcomeCentre/aeon_mecha.git
```

Then from the interpreter calling `import aeon` will report the following error trace:
```
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "C:\dev\python\aeon\.venv\Lib\site-packages\aeon\__init__.py", line 13, in <module>
    from aeon.io.api import load
ModuleNotFoundError: No module named 'aeon.io'
```

### Possible causes

Looking into the contents of the `aeon` package inside the local `site-packages` folder reveals that indeed none of the submodules have been included. My suspicion is this relates to how the wheels are being built from the `pyproject.toml` file, specifically the way the packages to be included are declared:

https://github.com/SainsburyWellcomeCentre/aeon_mecha/blob/25cc4b724d0451bf1171a70791bc3a3ed7d8d698/pyproject.toml#L79

I believe that we may need some wildcards to correctly include all subfolders. In a different project I use the following declaration:

https://github.com/harp-tech/harp-python/blob/c90387efdbb4d762073201b763025dd49d5d3117/pyproject.toml#L57
```
[tool.setuptools.packages.find]
include = ["harp*"]
```

The wildcard `*` may be necessary to allow building the wheels with all sub modules.

glopesdev:
As explained in the [package discovery docs for setuptools](https://setuptools.pypa.io/en/latest/userguide/package_discovery.html) each submodule needs to be either explicitly added, or the `find` feature should be used to automatically discover all relevant packages.

This PR replaces the `setuptools` declaration in `pyproject.toml` to automatically include all submodules inside the `aeon` folder when building the wheel, e.g. when installing the repo from `pip` in non-editable mode.

We can consider whether we want to further split the package into different sub-wheels or features for the datajoint pipeline, or other optional functionality, but this can be done in a separate discussion / PR.

Fixes #375

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -75,8 +75,8 @@ Repository = "https://github.com/sainsburyWellcomeCentre/aeon_mecha"
 Documentation = "https://sainsburywellcomecentre.github.io/aeon_docs/"
 DataJoint = "https://docs.datajoint.org/"
 
-[tool.setuptools]
-packages = ["aeon"]
+[tool.setuptools.packages.find]
+include = ["aeon*"]
 
 [tool.black]
 line-length = 108


pyproject.toml
```diff
@@ -75,8 +75,8 @@ Repository = "https://github.com/sainsburyWellcomeCentre/aeon_mecha"
 Documentation = "https://sainsburywellcomecentre.github.io/aeon_docs/"
 DataJoint = "https://docs.datajoint.org/"
 
-[tool.setuptools]
-packages = ["aeon"]
+[tool.setuptools.packages.find]
```
jkbhagatio, line 78: Apparently this pattern can also be used. Not sure if one is preferred, / if the * wildcard is necessary?

```
[tool.setuptools.packages.find]
where = ["aeon"]
```

lochhh:
```
[tool.setuptools.packages.find]
where = ["src"]  # list of folders that contain the packages (["."] by default)
include = ["my_package*"]  # package names should match these glob patterns (["*"] by default)
exclude = ["my_package.tests*"]  # exclude packages matching these glob patterns (empty by default)
namespaces = false  # to disable scanning PEP 420 namespaces (true by default)
```

[Ref](https://setuptools.pypa.io/en/latest/userguide/pyproject_config.html#:~:text=%5Btool.setuptools.packages,true%20by%20default)

glopesdev: I agree `where` probably makes the most sense in this case.

jkbhagatio (approved):

lochhh (approved):

pyproject.toml
glopesdev: Actually as @lochhh pointed out, `where` seems to move the working directory from which to look for packages. However, if we do this, then we lose the actual package folder structure, i.e. the package will not start with `aeon`.

Given this I still think that `include` is most appropriate for now.

lochhh: if we go for `include` we should also specify `exclude = ["docs"]` to exclude `aeon/dj_pipeline/docs`

PR #380, opened 2024-07-16
fix(Pose): remove `config_file` from class property

ttngu207:
Follow up on the comments on PR #372 
Remove `config_file` from class property and update other methods accordingly


jkbhagatio (approved):

ttngu207:
Replaced by #384

PR #397, opened 2024-08-22
Add table & reader for `Environment Active Configuration`

### LINKED ISSUE #391: Add reader for 'Environment_ActiveConfiguration' stream

ttngu207:
Fix #391

### DIFF
--- a/aeon/dj_pipeline/acquisition.py
+++ b/aeon/dj_pipeline/acquisition.py
@@ -588,7 +588,7 @@ class EnvironmentActiveConfiguration(dj.Imported):
             ),
         )
         device = devices_schema.Environment
-        stream_reader = device.EnvActiveConfiguration  # expecting columns: time, name, value
+        stream_reader = device.EnvironmentActiveConfiguration  # expecting columns: time, name, value
         stream_data = io_api.load(
             root=data_dirs,
             reader=stream_reader,
--- a/aeon/dj_pipeline/acquisition.py
+++ b/aeon/dj_pipeline/acquisition.py
@@ -563,6 +563,47 @@ class Environment(dj.Imported):
             )
 
 
+@schema
+class EnvironmentActiveConfig(dj.Imported):
+    definition = """  # Environment Active Configuration
+    -> Chunk
+    """
+
+    class Name(dj.Part):
+        definition = """
+        -> master
+        time: datetime(6)  # time when the configuration is applied to the environment
+        ---
+        name: varchar(32)  # name of the environment configuration
+        value: longblob    # dictionary of the configuration
+        """
+
+    def make(self, key):
+        chunk_start, chunk_end = (Chunk & key).fetch1("chunk_start", "chunk_end")
+        data_dirs = Experiment.get_data_directories(key)
+        devices_schema = getattr(
+            aeon_schemas,
+            (Experiment.DevicesSchema & {"experiment_name": key["experiment_name"]}).fetch1(
+                "devices_schema_name"
+            ),
+        )
+        device = devices_schema.Environment
+        stream_reader = device.ActiveConfiguration  # expecting columns: time, name, value
+        stream_data = io_api.load(
+            root=data_dirs,
+            reader=stream_reader,
+            start=pd.Timestamp(chunk_start),
+            end=pd.Timestamp(chunk_end),
+        )
+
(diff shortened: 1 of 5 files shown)


jkbhagatio (approved):

aeon/schema/social_03.py
```diff
@@ -6,3 +8,19 @@ class Pose(Stream):
 
     def __init__(self, path):
         super().__init__(_reader.Pose(f"{path}_202_*"))
+
+
+class EnvActiveConfigReader(_reader.JsonList):
```
glopesdev, line 13: Pulling out specific attributes from the JSON into columns might be a common enough operation that we could provide it at the level of the general reader, i.e. provide a `columns` property to specify which values to pull out from each record.

glopesdev, line 13:
```suggestion
class ActiveConfigurationReader(_reader.JsonList):
```

The names of the reader and stream don't line up, since one is `EnvActiveConfig` and the other is `ActiveConfiguration`. Why not just rename the reader as `ActiveConfigurationReader`? I think the `Env` prefix might be unnecessary in this context.

aeon/dj_pipeline/acquisition.py
```diff
@@ -563,6 +563,47 @@ def make(self, key):
             )
 
 
+@schema
+class EnvironmentActiveConfig(dj.Imported):
```
glopesdev, line 567: Would be nice if we can agree on whether to use `Config` or `Configuration` as a common suffix for all classes dealing with these files. I think I am fine either way.

aeon/schema/social_03.py
ttngu207: Agreed, I'll rename

ttngu207: For example, we can do something like this:

```
class JsonList(Reader):
    """Extracts data from json list (.jsonl) files, where the key "seconds"
    stores the Aeon timestamp, in seconds.
    """

    def __init__(self, pattern, columns=None, col_tuples=[], extension="jsonl"):
        super().__init__(pattern, columns, extension)
        self.col_tuples = col_tuples

    def read(self, file):
        """Reads data from the specified jsonl file."""
        with open(file, "r") as f:
            df = pd.read_json(f, lines=True)
        df.set_index("seconds", inplace=True)
        for new_col, from_attr in self.col_tuples:
            df[new_col] = df.apply(lambda x: x[from_attr])
        return df
```

jkbhagatio: In this case I actually like the `Env` prefix because it's more readable

aeon/schema/social_03.py
```diff
@@ -6,3 +8,19 @@ class Pose(Stream):
...
+    def __init__(self, pattern):
+        super().__init__(pattern)
+
+    def read(self, file):
+        data = super().read(file)
+        data["name"] = data["value"].apply(lambda x: x["name"])
+        return data
+
+
+class ActiveConfiguration(Stream):
```
jkbhagatio, line 23: would actually add `Env` here too

aeon/schema/social_03.py
ttngu207: (with better naming... `col_tuples` is a horrible name)

jkbhagatio: @glopesdev I don't have a strong preference on whether the pulling out of the column names occurs (whether in `JsonList` or in the lower-level Reader), so I guess you can make a final call here and then we can just go with that?

ttngu207: I've renamed to `EnvActiveConfigurationReader` - keeping the `Env` here to communicate more explicitly that this is for the Environment Active Configuration, just `ActiveConfiguration` may be too vague or too general

aeon/schema/social_03.py
```diff
@@ -6,3 +8,9 @@ class Pose(Stream):
 
     def __init__(self, path):
         super().__init__(_reader.Pose(f"{path}_202_*"))
+
+
+class EnvActiveConfiguration(Stream):
```
glopesdev, line 13: All looks great, I would just rename this stream to be called `EnvironmentActiveConfiguration` rather than the shorthand version, but not critical either way.

ttngu207: Done

ttngu207:
@jkbhagatio should be ready to merge now

Issue #396, opened 2024-08-21
Optimization of database storage


ttngu207:
Data size stored in mysql server (on `aeon-db2`) is approaching 400GB
This will hit the storage size limit of the `aeon-db2` server, at which point we will need to request for a disk size increase.

I'd like to discuss potential solutions to better engineer/optimize for storage space in our next data architecture meeting.

The bulk of the storage is in the `streams` schema, where we extract and store the data from relevant streams. Among these, the majority of the storage is at `UndergroundFeederEncoder` (the wheel data) (~100GB) and `SpinnakerVideoSourceVideo` (~80GB)

ttngu207:
I'm sharing here the current size (in GB) of each schema in the `aeon-db2` database

```
{'aeon_streams': 186.279919616,
 'aeon_test_streams': 13.5774208,
 'aeon_test_block_analysis': 0.000180224,
 'aeon_archived_exp02_acquisition': 42.648174592,
 'aeon_test_tracking': 0.216285184,
 'aeon_worker': 0.092962816,
 'aeon_tracking': 32.597950464,
 'aeon_test_worker': 0.001687552,
 'aeon_archived_exp01_analysis': 0.010092544,
 'aeon_test_acquisition': 0.010010624,
 'aeon_analysis': 0.00188416,
 'aeon_archived_exp02_subject': 0.000835584,
 'aeon_archived_exp02_worker': 0.000131072,
 'aeon_report': 4.9152e-05,
 'aeon_block_analysis': 23.938367488,
 'aeon_archived_exp02_report': 0.002752512,
 'aeon_device': 3.2768e-05,
 'u_thinh_aeonfix': 0.00475136,
 'aeon_subject': 0.000638976,
 'aeon_archived_exp02_streams': 0.000507904,
 'aeon_archived_exp01_acquisition': 0.10944512,
 'aeon_archived_exp02_qc': 24.861868032,
 'aeon_test_subject': 0.000278528,
 'aeon_archived_exp02_analysis': 4.933648384,
 'aeon_lab': 0.000442368,
 'aeon_archived_exp01_lab': 0.000425984,
 'aeon_test_qc': 9.8304e-05,
 'aeon_archived_exp01_subject': 0.000294912,
 'aeon_archived_exp01_report': 0.041025536,
 'aeon_archived_exp01_qc': 26.919780352,
 'aeon_test_analysis': 0.304234496,
[...]

PR #401, opened 2024-09-06
Allow forwarding of load function kwargs to reader

glopesdev:
To allow more flexible parameterization of data readers, we introduce support for `kwargs` forwarding in the main `load` function. The idea is to have these extra optional arguments, if present, be forwarded to the provided `reader` object when loading raw data chunks, as a way of further parameterizing the reader functionality on a "per-load" basis.

The motivation for this is related to #396 and the need to simultaneously downsample wheel data for ingestion and analysis, and access the raw data for debugging and testing.

### DIFF
--- a/aeon/io/api.py
+++ b/aeon/io/api.py
@@ -60,7 +60,7 @@ def _empty(columns):
     return pd.DataFrame(columns=columns, index=pd.DatetimeIndex([], name="time"))
 
 
-def load(root, reader, start=None, end=None, time=None, tolerance=None, epoch=None):
+def load(root, reader, start=None, end=None, time=None, tolerance=None, epoch=None, **kwargs):
     """Extracts chunk data from the root path of an Aeon dataset using the specified data stream
     reader. A subset of the data can be loaded by specifying an optional time range, or a list
     of timestamps used to index the data on file. Returned data will be sorted chronologically.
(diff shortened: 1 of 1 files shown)


jkbhagatio:
Hi @glopesdev , this makes sense to me conceptually, but I might suggest using a specific arg name (that will be a dict) instead of kwargs, to differentiate args that will be passed to load alone vs. args that will be passed to a read via load.

I think this also implies base class Reader's read should accept this new dict param, that i guess by default can be set to None or an empty dict?

glopesdev:
> Hi @glopesdev , this makes sense to me conceptually, but I might suggest using a specific arg name (that will be a dict) instead of kwargs, to differentiate args that will be passed to load alone vs. args that will be passed to a read via load.

Since we don't have `kwargs` in the main `load` function, I am assuming your concern here is a possible clash between the existing named arguments in `load` and a keyword argument in `read`? Given those arguments are all related to slicing and chunked loading I don't see how this should ever be required. We can address this nevertheless by explicit documentation.

The design I was going for is to use `kwargs` similar to [`pandas.DataFrame.plot`](https://pandas.pydata.org/pandas-docs/stable/reference/api/pandas.DataFrame.plot.html), where they are passed directly to the selected plotting method. This is very similar to our case, where we use `kwargs` to parameterize the `reader` argument which is the main extensibility point of `load`.

The disadvantage of the explicit `kwargs` is it creates heavier syntax on the `load` call and especially now that [ruff recommends against creating dictionaries using named arguments, and rather through explicit string literals](https://docs.astral.sh/ruff/rules/unnecessary-collection-call/), it makes the two syntaxes very different, e.g. compare below:

#### Implicit `kwargs`

```python
data = aeon.load(root, exp02.Patch2.Encoder, epoch=behavior_epoch, downsample=False)
```

#### Explicit `kwargs`

```python
data = aeon.load(root, exp02.Patch2.Encoder, epoch=behavior_epoch, reader_kwargs = { "downsample": False })
```

jkbhagatio:
Hey @glopesdev yes I had confused something in my mind, but we're all in favor of this, so will merge

jkbhagatio (approved):

PR #398, opened 2024-08-22
Update logic to retrieve threshold-specific pellet delivery timestamps

ttngu207:
Fix #382 
Fix #403 

High level logic for block detection
On a per-chunk basis, check for the presence of new block, insert into Block table.
1. Find the 0s in pellet_ct (these are times when the pellet count reset - i.e. new block)
2. Remove any double 0s (0s within 1 second of each other) (pick the first 0)
3. Calculate block end_times (use due_time) and durations
4. Insert into Block table

PR #408, opened 2024-09-11
Fix block detection, using `due_time`

### LINKED ISSUE #403: Use Block Due Time from `Environment_BlockState` in ingestion
Currently for analysis we try to find the start and end of blocks by looking for when the pellet count resets to 0 in the block state file.

We haven't been using the fact that the Block Due Time is also saved in this file as the last column. 

We can use this as an exact measure of when each block ends, which will simplify ingestion of block detection a fair bit, since now we don't even need to look across chunks.

ttngu207:
Fix #403 

High level logic for block detection
On a per-chunk basis, check for the presence of new block, insert into Block table.
      1. Find the 0s in `pellet_ct` (these are times when the pellet count reset - i.e. new block)
      2. Remove any double 0s (0s within 1 second of each other) (pick the first 0)
      3. Calculate block end_times (use due_time) and durations
      4. Insert into Block table

### DIFF
--- a/aeon/dj_pipeline/analysis/block_analysis.py
+++ b/aeon/dj_pipeline/analysis/block_analysis.py
@@ -217,10 +217,10 @@ class BlockAnalysis(dj.Computed):
                 encoder_df, maintenance_period, block_end, dropna=True
             )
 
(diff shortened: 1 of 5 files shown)


ttngu207:
Closing this PR, include changes in PR #398

PR #438, opened 2024-10-18
New `ingestion_schemas` + MANY minor fixes and improvements

### LINKED ISSUE #427: Add table for foraging bouts in BlockAnalysis
@ttngu207 initially proposed this:

https://github.com/SainsburyWellcomeCentre/aeon_mecha/pull/410#issuecomment-2357203253

and I suggested just removing saving 'bout_duration' in the table as its trivial for the user to compute given the start and end

ttngu207:
This PR includes a few major new logic/features and many minor fixes (mostly for BlockAnalysis)
1. Use `ingestion_schemas` - i.e. a separate set of schemas for DJ ingestion (specialized `Encoder` reader and `Video` reader)
    - Encoder reader with default downsampling to 50Hz
    - Video reader with only the `hw_timestamp` column
2. Update `streams.py` DJ schema accordingly (mainly to drop extra columns when reading Video data)
3. Add special handling to ingest fullpose data for `social0.2`
4. Add a new tracking table: `BlobPosition` to read and store blob position tracking when SLEAP data is not available
5. Fixes & improvements for BlockAnalysis
    - Bugfix incorrect extraction of `subject_in_patch` times
    - `fetch_stream` function rounds times to microseconds (mysql precision is to microseconds only)
    - BlockAnalysis use `BlobPosition` when `SLEAPTracking` is not available
    - improve logic to search for chunks in a given block
    - `BlockDetection` - when double 0s are found, use the first 0s (instead of the 2nd one)
    - Add `BlockForaging` computed table
    
Fix #427 
Replace #437


aeon/dj_pipeline/analysis/block_analysis.py
```diff
@@ -181,17 +182,27 @@ def make(self, key):
...
+                raise ValueError(
+                    f"BlockAnalysis Not Ready - SLEAPTracking (and BlobPosition) not yet fully ingested for block: {key}. Skipping (to retry later)..."
+                )
+            else:
+                use_blob_position = True
+
         # Patch data - TriggerPellet, DepletionState, Encoder (distancetravelled)
         # For wheel data, downsample to 10Hz
-        final_encoder_fs = 10
+        final_encoder_hz = 10
```
jkbhagatio, line 204: I think we actually want this to be 50 hz, not 10 hz

aeon/dj_pipeline/analysis/block_analysis.py
```diff
@@ -233,51 +244,51 @@ def make(self, key):
...
-                    {
-                        "note_timestamp": datetime.utcnow(),
-                        "note_type": "Multiple patch rates",
-                        "note": f"Found multiple patch rates for block {key} - patch: {patch_name} - rates: {depletion_state_df.rate.unique()}",
-                    }
-                )
+            if encoder_df.empty:
+                encoder_df["distance_travelled"] = 0
+            else:
+                encoder_df["distance_travelled"] = -1 * analysis_utils.distancetravelled(encoder_df.angle)
```
jkbhagatio, line 254: maybe add a comment saying something like -1 is for placement of magnetic encoder, where wheel movement actually decreases encoder value?

aeon/dj_pipeline/analysis/block_analysis.py
```diff
@@ -288,27 +299,50 @@ def make(self, key):
             & f'chunk_start <= "{chunk_keys[-1]["chunk_start"]}"'
         )[:block_start]
         subject_visits_df = subject_visits_df[subject_visits_df.region == "Environment"]
+        subject_visits_df = subject_visits_df[~subject_visits_df.id.str.contains("Test", case=False)]
```
jkbhagatio, line 302: sometimes we use other, non "Test" subjects as test subjects. Maybe the check should be, if the subject does not begin with 'baa' (can str.lower to check for regardless of case) ?

aeon/io/reader.py
```diff
@@ -76,9 +75,12 @@ def read(self, file):
         if self.columns is not None and payloadshape[1] < len(self.columns):
             data = pd.DataFrame(payload, index=seconds, columns=self.columns[: payloadshape[1]])
             data[self.columns[payloadshape[1] :]] = math.nan
-            return data
         else:
-            return pd.DataFrame(payload, index=seconds, columns=self.columns)
+            data = pd.DataFrame(payload, index=seconds, columns=self.columns)
+
+        # remove rows where the index is zero (why? corrupted data in harp files?)
+        data = data[data.index != 0]
```
jkbhagatio, line 82: This I guess will be fixed in the next PR targeting a new Encoder reader in ingestion_schemas.py ?

aeon/schema/social_02.py
```diff
@@ -48,6 +48,12 @@ def __init__(self, path):
         super().__init__(_reader.Pose(f"{path}_test-node1*"))
 
 
+class Pose03(Stream):
```
jkbhagatio, line 51: made add a comment that this is necessary due to changing registers for the pose streams for social02 in particular? And that 03 corresponds to the fact that this is because this pattern is what we're going with for social03 and moving forward? Or call this class something else?

jkbhagatio: Or just import `Pose` from social_03.py as `Pose03`? I think I like something like this actually

jkbhagatio (commented):
I've left a bunch of small comments here.

A general point - maybe add top-level module docstrings for all these modules? Unless this will be done by @MilagrosMarin in #443 ?

aeon/dj_pipeline/__init__.py
```diff
@@ -30,11 +31,17 @@ def dict_to_uuid(key) -> uuid.UUID:
...
+def fetch_stream(query, drop_pk=True, round_microseconds=True):
     """Fetches data from a Stream table based on a query and returns it as a DataFrame.
 
     Provided a query containing data from a Stream table,
     fetch and aggregate the data into one DataFrame indexed by "time"
+
+    Args:
+        query (datajoint.Query): A query object containing data from a Stream table
+        drop_pk (bool, optional): Drop primary key columns. Defaults to True.
+        round_microseconds (bool, optional): Round timestamps to microseconds. Defaults to False.
```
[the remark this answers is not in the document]
jkbhagatio: For me this is ok?

ttngu207:
> I've left a bunch of small comments here.
> 
> A general point - maybe add top-level module docstrings for all these modules? Unless this will be done by @MilagrosMarin in #443 ?

Yes, this will be address when we apply Ruff checks and fixes

aeon/dj_pipeline/analysis/block_analysis.py
ttngu207: The Encoder data is ingested at 50Hz, this is at the BlockAnalysis step, do we also want 50Hz here (or further downsample to 10?)

aeon/dj_pipeline/analysis/block_analysis.py
```diff
@@ -233,51 +244,51 @@ def make(self, key):
...
+                    )
 
-            encoder_fs = (
-                1 / encoder_df.index.to_series().diff().dt.total_seconds().median()
-            )  # mean or median?
-            wheel_downsampling_factor = int(encoder_fs / final_encoder_fs)
+                patch_rate = depletion_state_df.rate.iloc[0]
+                patch_offset = depletion_state_df.offset.iloc[0]
+                # handles patch rate value being INF
+                patch_rate = 999999999 if np.isinf(patch_rate) else patch_rate
```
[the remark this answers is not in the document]
ttngu207: Yes, it's due to MySQL `float` doesn't handle `INF` well - we can convert to `NaN` but that would lose the intended meaning of `INF` here

aeon/schema/social_02.py
ttngu207: This is primarily due to full-pose data for social0.2 has a different pattern than that of the non-fullpose SLEAP data.
I don't think you can do `from social_03 import Pose as Pose03`, the way the the Devices/Streams being instantiated is that it's using the class name, so aliasing wouldn't work.

aeon/dj_pipeline/__init__.py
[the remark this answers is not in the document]
ttngu207: fixed

aeon/dj_pipeline/analysis/block_analysis.py
ttngu207: done

aeon/dj_pipeline/analysis/block_analysis.py
ttngu207: I see. Is `startswith("baa")` reliable, future-proof? Perhaps better logic is to also cross check with the Subjects to be manually specified by users for a particular experiment

aeon/dj_pipeline/analysis/block_analysis.py
jkbhagatio: yes we want 50 hz here too, thanks!

aeon/dj_pipeline/analysis/block_analysis.py
jkbhagatio: Yes, your suggestion would be a better check.

aeon/schema/social_02.py
jkbhagatio: ah I see, then just some documentation is ok

glopesdev:
This PR also fixed unexpected zero timestamps in Harp files (https://github.com/SainsburyWellcomeCentre/aeon_mecha/pull/438/commits/cb90843271f2b4824093f4c773c92c305092274b), e.g.:

```
/ceph/aeon/aeon/data/raw/AEON4/social0.3/2024-06-26T19-33-32/Patch1/Patch1_90_2024-06-28T23-00-00.bin
```

PR #458, opened 2024-12-12
add in_patch_rfid_timestamps in `BlockSubjectAnalysis`

### LINKED ISSUE #457: Add `in_patch_rfid_timestamps` to `BlockSubjectAnalysis.Patch`
The `BlockSubjectAnalysis.Patch` table already has a column `in_patch_timestamps` to store timestamps when a subject is at a specific patch.
However, this is not from RFID but is computed based on SLEAP tracking data.

We would like to, in addition to this, have another column that has the data coming from the RFID stream

ttngu207:
Fix #457


jkbhagatio (approved):
lgtm!

ttngu207:
Update script `update_in_patch_rfid_timestamps.py` executed.
All `BlockSubjectAnalysis` should have `in_patch_rfid_timestamps` extracted/added now.

PR #485, opened 2025-06-24
Refactor clean_swaps

lochhh:
## Fill early frames
In #484 the lines to fill early frames (before the first complete frame) were removed from `clean_swaps`. This means that if the input df starts with only one animal being detected for the first 1000 frames, these frames will be dropped (NaN) in the cleaned df. This PR reintroduces "filling early frames". 

## Merging
When merging the cleaned df with the original df any single-detection swaps will be dropped because we merge based on "identity_name":
https://github.com/SainsburyWellcomeCentre/aeon_mecha/blob/e94ce6c6ada739096cbbac75d2eea1561c9f2a2a/aeon/dj_pipeline/utils/tracking_utils.py#L243-L248
This PR fixes this by using a right merge instead, since the cleaned df will have at least as many rows as the raw df. 

## Vote count
In the parts below, depending on which point is closer, we may or may not swap the IDs. When we do swap, why do we still increment `votes_same` rather than `votes_swap`? 
https://github.com/SainsburyWellcomeCentre/aeon_mecha/blob/e94ce6c6ada739096cbbac75d2eea1561c9f2a2a/aeon/dj_pipeline/utils/tracking_utils.py#L187-L192
https://github.com/SainsburyWellcomeCentre/aeon_mecha/blob/e94ce6c6ada739096cbbac75d2eea1561c9f2a2a/aeon/dj_pipeline/utils/tracking_utils.py#L205-L214
This PR [changes](https://github.com/SainsburyWellcomeCentre/aeon_mecha/blob/eec05c12eaecf73f7cc4b21cf5a7420f8c5eae89/aeon/dj_pipeline/utils/tracking_utils.py#L163-L172) this by determining if a swap occurred and updating the vote count (`votes_same` or `votes_swap`) accordingly.

## Others
Other smaller changes include:
- adding helper functions to reduce repetition
- defining constants to replace magic numbers
- using google docstring style

### DIFF
--- a/aeon/dj_pipeline/utils/tracking_utils.py
+++ b/aeon/dj_pipeline/utils/tracking_utils.py
@@ -4,9 +4,68 @@ import numpy as np
 import pandas as pd
 from tqdm import tqdm
 
+NUM_IDS = 2  # Number of identities expected in the tracking data
+MAX_DIST_BETWEEN_FRAMES = 90  # Maximum distance between frames
+MAX_INTER_SUBJ_DIST = 100  # Maximum distance between two subjects
+MAX_SWAP_COST = 100  # Maximum cost distance to consider a swap
+
+
+def filter_valid_points(df: pd.DataFrame, region_df: pd.DataFrame) -> pd.DataFrame:
+    """Filter points in the DataFrame that are within the arena or nest region.
+
+    Args:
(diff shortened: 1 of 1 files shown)


anayapouget:
Thanks for the work @lochhh this looks great. I was wondering whether we should keep the tqdm bit? It was originally included for local testing but seems unecessary to have anymore

ttngu207 (approved):

PR #520, opened 2025-12-30
Spikeinterface gui integration

esutlie:
This pull request introduces a comprehensive workflow for manual spike sorting curation using the SpikeInterface GUI, and integrates this workflow with the DataJoint pipeline. It adds two new scripts for launching the GUI and saving curation results, and updates the pipeline logic to support curated and raw spike sorting sessions. The changes ensure curated results are properly tracked, saved, and can be designated as official or reverted as needed.

**New manual curation workflow scripts:**

* Added `launch_si_gui.py` for launching the SpikeInterface GUI with support for loading parent curations and handling user warnings about overwriting or resuming curation state.
* Added `save_curation.py` for saving manual curation results to the database, designating official curations, and restoring raw (uncurated) sorting. Includes robust file handling and user guidance.

**Pipeline integration and logic updates:**

* Updated `spike_sorting.py` to automatically detect and use curated analyzers if an official curation exists, falling back to raw results otherwise. This includes lazy importing to avoid circular dependencies and logging for transparency.
* Ensured `curation_id` is tracked in the `SortedSpikes` table and used to select the appropriate analyzer directory throughout downstream tables, maintaining consistency between raw and curated sessions.

### DIFF
--- a/aeon/dj_pipeline/docs/spike_sorting_curation_specifications.md
+++ b/aeon/dj_pipeline/docs/spike_sorting_curation_specifications.md
@@ -92,6 +92,8 @@ The table includes several important attributes. The `curation_datetime` field s
 
 The `ManualCuration.File` part table stores the file path information for each curation. It includes a `file_name` field storing the filename (e.g., "curation_data_id1.json") and a `file` field that uses DataJoint's filepath storage to reference the actual curation JSON file on disk. This separation allows the system to track file locations while keeping the main table focused on curation metadata.
 
+When an official curation is applied (via `ApplyOfficialCuration`), the system also stores the path to the curated analyzer directory in `ManualCuration.File` with `file_name="curation_applied_analyzer"`. This allows the system to fetch the applied analyzer directory path from the database instead of relying on naming conventions, making the system more maintainable and flexible.
+
 #### `OfficialCuration` (Manual Table)
 
 The `OfficialCuration` table designates which manual curation should be considered the "official" version for a given session. This table has a simple structure: its primary key inherits from `spike_sorting.SortedSpikes`, ensuring a one-to-one relationship with sorted spike data, and it contains a single attribute `curation_id` that references the `ManualCuration` table. This design enforces the constraint that only one curation can be official at a time for any given sorting session.
@@ -151,7 +153,7 @@ The sorting root directory is retrieved using the `get_sorting_root_dir()` funct
 
 The output directory for a specific sorting session is stored in the `PreProcessing` table's `sorting_output_dir` field as a relative path. The system retrieves this path using `sorting_root_dir / (PreProcessing & key).fetch1("sorting_output_dir")`, combining the root directory with the session-specific relative path. This design allows the system to work with data stored in various locations while maintaining a consistent interface.
 
-The analyzer directories follow a predictable pattern. The raw analyzer is always located at `output_dir / "sorting_analyzer"`, while curated analyzers are stored at `output_dir / f"sorting_analyzer_curated_id{curation_id}"`. This naming convention makes it easy to identify which analyzer is which and allows the system to automatically construct paths based on the `curation_id` stored in the database.
+The analyzer directories follow a predictable pattern. The raw analyzer is always located at `output_dir / "sorting_analyzer"`, while curated analyzers are stored at `output_dir / f"sorting_analyzer_curated_id{curation_id}"`. When an official curation is applied, the system stores the curated analyzer directory path in `ManualCuration.File` with `file_name="curation_applied_analyzer"`, allowing the system to fetch the path from the database instead of constructing it from a naming convention. This provides a single source of truth and makes the system more maintainable.
 
 ### File Lifecycle and Naming Conventions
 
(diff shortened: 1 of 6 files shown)


aeon/dj_pipeline/spike_sorting.py
```diff
@@ -619,8 +619,32 @@ def make(self, key):
                 & key
         )
 
-        # Get sorter method and create output directory.
+        # Check if there's an official curation for this session
+        # If so, use the curated analyzer; otherwise use the raw analyzer
+        # Use lazy import to avoid circular dependency
+        curation_id = -1
         analyzer_output_dir = output_dir / "sorting_analyzer"
+        try:
```
esutlie, line 627: This section uses importlib to try to avoid circular import logic when accessing the curation schema tables. If there is a more elegant way to handle this, this section could definitely be improved.

aeon/dj_pipeline/scripts/launch_si_gui.py
```diff
@@ -0,0 +1,196 @@
...
+"""
+
+from pathlib import Path
+from typing import Dict, Any, Optional
+import spikeinterface as si
+import shutil
+from aeon.dj_pipeline import spike_sorting, spike_sorting_curation
+
+
+def launch_si_gui(
```
ttngu207, line 35: Instead of in a separate script (launch_si_gui.py), I'd suggest that this function should be a helper function in the `spike_sorting_curation.py` module. 
You can still keep this `launch_is_gui.py` script, only the `if __name__ == `__main__`:` section, or not, set up another entrypoint instead.

Same with the `save_curation.py` script, should be helper function in `spike_sorting_curation.py` module

aeon/dj_pipeline/scripts/launch_si_gui.py
```diff
@@ -0,0 +1,196 @@
...
+        block_start = datetime.fromisoformat(block_start.replace(" ", "T"))
+    if isinstance(block_end, str):
+        block_end = datetime.fromisoformat(block_end.replace(" ", "T"))
+
+    start_str = block_start.strftime("%Y-%m-%dT%H-%M-%S")
+    end_str = block_end.strftime("%Y-%m-%dT%H-%M-%S")
+
+    # Construct path to sorting_analyzer directory
+    # Structure: experiment_name / ephys_blocks / {start}_{end} / electrode_group / {method}_{paramset} / sorting_analyzer
+    analyzer_dir = (
```
ttngu207: Instead of constructing the path from scratch, for robustness, fetch `sorting_output_dir` from `PreProcessing` and build the full path from there.

ttngu207: You can also verify here that an entry corresponding to this `key` must exist in `PostProcessing` (otherwise there won't be a sorting_analyzer)

aeon/dj_pipeline/docs/spike_sorting_curation_specifications.md
ttngu207, line 1: This specs is very good, but also at the same time way too long (for a human, but perhaps just fine for LLMs to read)

aeon/dj_pipeline/spike_sorting.py
ttngu207: You can call the normal import within this `make` function - that would avoid the circular import issue. It's not ideal but it should work. Using `importlib` is fine but may be a bit overkill

aeon/dj_pipeline/spike_sorting.py
```diff
@@ -756,8 +782,19 @@ def make(self, key):
                 & key
         )
 
-        # Get sorter method and create output directory.
-        analyzer_output_dir = output_dir / "sorting_analyzer"
+        # Get curation_id from SortedSpikes to determine which analyzer to use
+        curation_id = (SortedSpikes & key).fetch1("curation_id")
+        if curation_id != -1:
+            analyzer_output_dir = (
+                output_dir / f"sorting_analyzer_curated_id{curation_id}"
```
ttngu207: Is it possible to fetch this from somewhere instead of having to rely on the naming convention (f"sorting_analyzer_curated_id{curation_id}") here? from `ManualCuration.File`?

aeon/dj_pipeline/spike_sorting.py
```diff
@@ -829,8 +866,19 @@ def make(self, key):
         sorting_root_dir = get_sorting_root_dir()
         output_dir = sorting_root_dir / (PreProcessing & key).fetch1("sorting_output_dir")
 
-        # Get sorter method and create output directory.
-        analyzer_output_dir = output_dir / "sorting_analyzer"
+        # Get curation_id from SortedSpikes to determine which analyzer to use
+        curation_id = (SortedSpikes & key).fetch1("curation_id")
+        if curation_id != -1:
+            analyzer_output_dir = (
+                output_dir / f"sorting_analyzer_curated_id{curation_id}"
```
ttngu207: Same comment as above, fetch if possible, instead of having to rely on naming convention: `f"sorting_analyzer_curated_id{curation_id}"`

aeon/dj_pipeline/spike_sorting_curation.py
```diff
@@ -69,6 +70,498 @@ class ApplyOfficialCuration(dj.Imported):
...
+            "This should not happen - downstream tables should be auto-deleted."
+        )
+        assert not (spike_sorting.SortingQuality & key), (
+            "SortingQuality entries still exist after SortedSpikes deletion. "
+            "This should not happen - downstream tables should be auto-deleted."
+        )
+        assert not (spike_sorting.SyncedSpikes & key), (
+            "SyncedSpikes entries still exist after SortedSpikes deletion. "
+            "This should not happen - downstream tables should be auto-deleted."
+        )
```
ttngu207: No need for this check, it's impossible for this to happen, thanks to the relational database

aeon/dj_pipeline/docs/spike_sorting_curation_specifications.md
ttngu207: Perhaps a mermaid diagram could help making this more human-friendly

aeon/dj_pipeline/scripts/launch_si_gui.py
```diff
@@ -0,0 +1,52 @@
...
+        parent_curation_id: Optional curation_id to base this curation on. If provided,
+            the curation_data.json file will be initialized from the specified curation.
+            If None, starts from the raw sorting results.
+    """
+    spike_sorting_curation.launch_spikeinterface_gui(key, parent_curation_id)
+
+
+if __name__ == "__main__":
+    # Example key - modify these values for your session
+    key = {
```
ttngu207, line 59: Is it expected for users to modify this script file and then run `python launch_is_gui.py` - if so, that's not quite a standard behavior for a script
1. if the expectation is for user to modify the script, this probably should be a notebook
2. or use normal user-input (sys.args) for users to specify parameters for this script (I can see this being quite hideous)
3. keep as is, but add more/bigger docstring top of file to make it extra obvious that the `key` should be updated accordingly.

PR #531, opened 2026-02-18
Fix timezone mismatch in epoch/chunk ingestion

esutlie:
## Summary
- `io_api.load()` returns UTC-aware timestamps (`pd.to_datetime(utc=True)`), but `ingest_epochs`, `ingest_chunks`, and `EpochConfig.make` parse epoch directory names with `datetime.strptime()` which produces timezone-naive datetimes
- Comparing the two raises `TypeError: can't compare offset-naive and offset-aware datetimes`
- Adds `.replace(tzinfo=datetime.timezone.utc)` to all 4 `strptime` calls in `acquisition.py`, plus a defensive guard for `epoch_end` fetched from the DB via DataJoint (which returns naive datetimes from MySQL)

## How it was found
Hit during first end-to-end test of `tn/core_pipeline` with real foragingABC data on HPC. Crashed at `Epoch.ingest_epochs` line 242: `min(previous_chunk_end, epoch_start)`.

## Test plan
- [x] Verified all `strptime` calls in `acquisition.py` are covered
- [ ] Re-run `setup_pipeline.py` on HPC with this fix (in progress)

### DIFF
--- a/aeon/dj_pipeline/acquisition.py
+++ b/aeon/dj_pipeline/acquisition.py
@@ -215,7 +215,9 @@ class Epoch(dj.Manual):
         for i, (_, chunk) in enumerate(all_chunks.iterrows()):
             chunk_rep_file = pathlib.Path(chunk.path)
             epoch_dir = pathlib.Path(chunk_rep_file.as_posix().split(device_name)[0])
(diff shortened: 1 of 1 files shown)


esutlie:
Actually there're more issues with this further down the ingestion. I'll open a new one with a more complete fix.

PR #534, opened 2026-02-24
Refine unit matching design (spec + schema changes)

ttngu207:
## Summary
- Add design spec for unit matching system (`SPEC_unit_matching.md`)
- Key design refinements over `es/ephys_revamp`:
  - `UnitMatchingMethod` as non-PK FK (one matching result per session)
  - `Matched` as Part of `UnitMatching` (not `UniversalUnit`) for clean cascade
  - `ChunkedSpikeTimes` as Part of `UnitMatching` with ownership convention + `unique index` enforcement
  - Drop standalone `ChunkedSpikeTimes` Computed table

### DIFF
--- a/aeon/dj_pipeline/docs/specs/SPEC_unit_matching.md
+++ b/aeon/dj_pipeline/docs/specs/SPEC_unit_matching.md
@@ -16,13 +16,13 @@
 
 ### Purpose
 
-Long-running AEON experiments record neural activity across multiple consecutive ephys blocks (sessions), each independently spike-sorted. The same neuron appears as a different `unit` ID in each session's sorting results. The unit matching system solves this by assigning a persistent **global unit** identity to neurons across sessions, enabling longitudinal analysis of single-neuron activity over days or weeks.
+Long-running AEON experiments record neural activity across multiple consecutive ephys blocks, each independently spike-sorted. The same neuron appears as a different `unit` ID in each block's sorting results. The unit matching system solves this by assigning a persistent **global unit** identity to neurons across blocks, enabling longitudinal analysis of single-neuron activity over days or weeks.
 
-The system works by exploiting temporal overlap between consecutive ephys blocks: when two blocks share a time window, the same neurons produce spikes in both blocks' sorted data. By comparing spike times in the overlap region, the system identifies which units across sessions correspond to the same neuron.
+The system works by exploiting temporal overlap between consecutive ephys blocks: when two blocks share a time window, the same neurons produce spikes in both blocks' sorted data. By comparing spike times in the overlap region, the system identifies which units across blocks correspond to the same neuron.
 
 ### Prerequisites
 
-Unit matching sits at the end of the spike sorting pipeline. A session must have completed the full chain before it is eligible for matching:
+Unit matching sits at the end of the spike sorting pipeline. An ephys block must have completed the full chain before it is eligible for matching:
 
 ```
 SortingTask → PreProcessing → SpikeSorting → PostProcessing
@@ -30,27 +30,27 @@ SortingTask → PreProcessing → SpikeSorting → PostProcessing
 → UnitMatching (this system)
 ```
 
-Specifically, `UnitMatching.key_source` requires that `ApplyOfficialCuration` exists for the session, ensuring only curated (quality-controlled) results enter the matching pipeline.
+Specifically, `UnitMatching.key_source` requires that `ApplyOfficialCuration` exists for the block, ensuring only curated (quality-controlled) results enter the matching pipeline.
 
 ### Key Concepts
 
 A **global unit** is a persistent neuron identity scoped to a single probe insertion (i.e., one physical probe in one subject). Global unit IDs are integers starting at 1, unique within a `(experiment_name, subject, insertion_number)` scope.
 
-A **matched unit** links a session-specific `SortedSpikes.Unit` to its assigned global unit. Each session unit maps to exactly one global unit. A global unit may be matched to units from multiple sessions (that's the whole point).
+A **matched unit** links a block-specific `SortedSpikes.Unit` to its assigned global unit. Each block's unit maps to exactly one global unit. A global unit may be matched to units from multiple blocks (that's the whole point).
 
 **Temporal overlap** between consecutive ephys blocks is the mechanism that enables matching. The overlap region contains spikes from the same neurons captured by both blocks' independent sorting runs. The matching algorithm compares spike times in this overlap window to identify corresponding units.
 
-The **ownership convention** determines which session "owns" the spike data for a given (global_unit, chunk) pair when multiple sessions cover the same chunk. This prevents duplicate rows in `Spikes`.
+The **ownership convention** determines which block "owns" the spike data for a given (global_unit, chunk) pair when multiple blocks cover the same chunk. This prevents duplicate rows in `Spikes`.
 
 ### Critical Constraint: Temporal Ordering
 
-Sessions **must** be processed in temporal order (by `block_start`). Session N compares its units against previously-matched sessions (1..N-1). Processing out of order produces incorrect global unit assignments.
+Ephys blocks **must** be processed in temporal order (by `block_start`). Block N compares its units against previously-matched blocks (1..N-1). Processing out of order produces incorrect global unit assignments.
 
 This constraint is enforced at two levels:
 
-1. **`key_source` gate**: Only yields the session with the earliest unprocessed `block_start` per insertion. Under `populate(reserve_jobs=True)`, a worker that finishes the earliest session causes the next `key_source` evaluation to advance to the next block. Workers processing different insertions can run in parallel; within an insertion, processing is serialized.
+1. **`key_source` gate**: Only yields the block with the earliest unprocessed `block_start` per insertion. Under `populate(reserve_jobs=True)`, a worker that finishes the earliest block causes the next `key_source` evaluation to advance to the next block. Workers processing different insertions can run in parallel; within an insertion, processing is serialized.
 
-2. **`make()` guard**: Before processing, verifies that no earlier eligible session (by `block_start`) for the same insertion remains unprocessed. This is a defense-in-depth check that catches direct `make(key)` calls bypassing `key_source`, or hypothetical bugs in the key_source logic.
+2. **`make()` guard**: Before processing, verifies that no earlier eligible block (by `block_start`) for the same insertion remains unprocessed. This is a defense-in-depth check that catches direct `make(key)` calls bypassing `key_source`, or hypothetical bugs in the key_source logic.
 
 ```python
 # key_source: only the earliest unprocessed block per insertion
@@ -67,10 +67,10 @@ return candidates * next_per_insertion & "block_start = next_start"
 earlier_eligible = eligible & insertion_key & f'block_start < "{key["block_start"]}"'
 unprocessed_earlier = earlier_eligible - self
 if unprocessed_earlier:
-    raise ValueError("Temporal ordering violation: earlier sessions not yet processed.")
+    raise ValueError("Temporal ordering violation: earlier blocks not yet processed.")
 ```
 
-**Interaction with `reserve_jobs=True`**: When Worker A reserves the earliest session for insertion I1, Worker B's `key_source` also returns that same session (it's not yet in `self`). But the job reservation system prevents B from reserving the already-reserved key. With no other candidates for I1, Worker B idles or processes a different insertion. When A finishes, the next `populate()` cycle advances to the next block.
+**Interaction with `reserve_jobs=True`**: When Worker A reserves the earliest block for insertion I1, Worker B's `key_source` also returns that same block (it's not yet in `self`). But the job reservation system prevents B from reserving the already-reserved key. With no other candidates for I1, Worker B idles or processes a different insertion. When A finishes, the next `populate()` cycle advances to the next block.
 
 ---
 
@@ -85,7 +85,7 @@ UnitMatchingMethod (Lookup)
     matching_method_description: varchar(1000)
 
 UnitMatching (Computed)
-    -> SyncedSpikes                         # inherits full session key
+    -> SyncedSpikes                         # inherits full block key
     ---
     -> UnitMatchingMethod                   # non-PK: which method was used
     execution_time: datetime
@@ -112,7 +112,7 @@ GlobalUnit (Manual)
     -> ephys.ProbeInsertion                 # experiment_name, subject, insertion_number
     global_unit: int                     # unique ID within this insertion
     ---
-    -> ephys.ProbeType.Electrode      # peak electrode (denormalized, updated each session)
+    -> ephys.ProbeType.Electrode      # peak electrode (denormalized, updated each block)
     global_unit_comment='': varchar(1000)
 ```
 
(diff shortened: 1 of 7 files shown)


aeon/dj_pipeline/docs/specs/SPEC_unit_matching.md
```diff
@@ -0,0 +1,345 @@
...
+```
+UnitMatchingMethod (Lookup)
+    matching_method: varchar(32)
+    ---
+    matching_method_description: varchar(1000)
+
+UnitMatching (Computed)
+    -> SyncedSpikes                         # inherits full session key
+    ---
+    -> UnitMatchingMethod                   # non-PK: which method was used
```
esutlie: How should UnitMatching identify which matching method to use?

aeon/dj_pipeline/docs/specs/SPEC_unit_matching.md
```diff
@@ -0,0 +1,345 @@
...
+
+A **matched unit** links a session-specific `SortedSpikes.Unit` to its assigned universal unit. Each session unit maps to exactly one universal unit. A universal unit may be matched to units from multiple sessions (that's the whole point).
+
+**Temporal overlap** between consecutive ephys blocks is the mechanism that enables matching. The overlap region contains spikes from the same neurons captured by both blocks' independent sorting runs. The matching algorithm compares spike times in this overlap window to identify corresponding units.
+
+The **ownership convention** determines which session "owns" the spike data for a given (universal_unit, chunk) pair when multiple sessions cover the same chunk. This prevents duplicate rows in `ChunkedSpikeTimes`.
+
+### Critical Constraint: Temporal Ordering
+
+Sessions **must** be processed in temporal order (by `block_start`). Session N compares its units against previously-matched sessions (1..N-1). Processing out of order produces incorrect universal unit assignments. This constraint is documented but enforced only by convention — there is no schema-level enforcement.
```
esutlie: Is this something that we want to accept? Is there any way to enforce that this table is populated serially? I don't think it's strictly necessary that the rows be processed in temporal order, the algorithm can handle merges, the rows just can't be processed in parallel.

aeon/dj_pipeline/docs/specs/SPEC_unit_matching.md
```diff
@@ -0,0 +1,345 @@
...
+    "ephys.ProbeInsertion" ||--|| UniversalUnit : "scopes"
+```
+
+### Design Notes
+
+- **`UnitMatchingMethod` as non-PK FK**: Only one matching result per session. The method is recorded for provenance but does not multiply the key space. This means one canonical set of universal units per insertion — no parallel method comparison. This is a deliberate simplification; if method comparison becomes needed, it can be added later by promoting the FK back to PK.
+
+- **`UniversalUnit` is Manual**: Populated programmatically by `UnitMatching.make()`, not by DataJoint's `populate()` machinery. This means it cannot be auto-cleared. Orphaned entries (those with no remaining `Matched` references) must be cleaned up explicitly during re-processing.
+
+- **`unique index` on `ChunkedSpikeTimes`**: The Part table PK includes the master's session key, so the same `(universal_unit, chunk_start)` from different sessions would be valid at the PK level. The unique index on `(universal_unit, chunk_start)` provides a schema-level guarantee that only one session can own a given (unit, chunk) pair, making the ownership convention a hard constraint rather than a soft convention.
```
esutlie: `universal_unit` is currently meant to be unique to the insertion, but with multiple probes implanted in the same mouse, you can have more than one valid `(universal_unit, chunk_start)`.

aeon/dj_pipeline/docs/specs/SPEC_unit_matching.md
```diff
@@ -0,0 +1,345 @@
...
+        -> UniversalUnit                    # universal_unit as dependent FK
+        match_confidence=null: float
+        match_comment='': varchar(1000)
+
+    UnitMatching.ChunkedSpikeTimes (Part)
+        -> master
+        -> UniversalUnit                    # adds universal_unit to PK
+        -> ephys.EphysChunk                 # adds chunk_start to PK
+        ---
+        spike_times: longblob               # float64 epoch seconds (UTC), HARP-synced
```
esutlie: Should this be `datetime64[ns]`?

esutlie (commented):
Looks great! Definitely an improvement. I left a few questions scattered throughout, but since it's just the spec doc we can merge whenever you want and I can integrate it into my  own spec docs.

ttngu207:
Re: @esutlie's [comment on temporal ordering](https://github.com/SainsburyWellcomeCentre/aeon_mecha/pull/534#discussion_r2850212445):

> Is this something that we want to accept? Is there any way to enforce that this table is populated serially?

This is now addressed in the second commit (38efa07). The spec has been updated with a two-level enforcement mechanism:

1. **`key_source` gate**: Only yields the session with the earliest unprocessed `block_start` per insertion (using `MIN(block_start)` aggregation). Under `populate(reserve_jobs=True)`, a worker that reserves the earliest session prevents other workers from grabbing later sessions for the same insertion — they find nothing to reserve and either idle or work on a different insertion. This naturally serializes processing per insertion while still allowing parallelism across insertions.

2. **`make()` guard**: Defensive check that raises `ValueError` if any earlier eligible session for the same insertion hasn't been processed. This catches direct `make(key)` calls that bypass `key_source`.

See the updated "Critical Constraint: Temporal Ordering" section in the spec for the full implementation details including the `key_source` code.

ttngu207:
Re: @esutlie's [comment on unique index scope](https://github.com/SainsburyWellcomeCentre/aeon_mecha/pull/534#discussion_r2850250347):

> `universal_unit` is currently meant to be unique to the insertion, but with multiple probes implanted in the same mouse, you can have more than one valid `(universal_unit, chunk_start)`.

Good catch — this was a bug. Fixed in 33c9550. The unique index is now scoped per-insertion:

```
unique index (experiment_name, subject, insertion_number, universal_unit, chunk_start)
```

Two insertions for the same subject can both have `universal_unit=1` in the same chunk without conflict.

ttngu207:
Re: @esutlie's [comment on matching method selection](https://github.com/SainsburyWellcomeCentre/aeon_mecha/pull/534#discussion_r2850204863):

> How should UnitMatching identify which matching method to use?

This is a deliberate design tradeoff documented in the "Design Decisions and Rationale" section of the spec. With `UnitMatchingMethod` as a non-PK FK (rather than part of the primary key), there's exactly one matching result per session. In practice, `make()` would use the sole entry in `UnitMatchingMethod` (currently `spike_time_overlap`) — effectively hardcoded.

The tradeoff: we get simpler downstream queries (no `matching_method` to carry through every join) and one canonical set of universal units per insertion, at the cost of not being able to compare multiple methods side-by-side. If method comparison becomes needed later, the FK can be promoted back to PK.

ttngu207:
Re: @esutlie's [comment on spike_times format](https://github.com/SainsburyWellcomeCentre/aeon_mecha/pull/534#discussion_r2850272259):

> Should this be `datetime64[ns]`?

Yes — fixed in 07228ae. `ChunkedSpikeTimes.spike_times` now stores `datetime64[ns]`, matching `SyncedSpikes.Unit`. The conversion to epoch seconds only happens internally during the matching algorithm for SpikeInterface compatibility.

ttngu207:
## Design update summary (latest commits)

Several refinements since the initial spec, driven by query UX review:

### Naming
- **`UniversalUnit` → `GlobalUnit`** (`universal_unit` → `global_unit`) — shorter, same meaning
- **Part table names**: `UnitMatching.Matched` → `UnitMatching.Unit`, `UnitMatching.ChunkedSpikeTimes` → `UnitMatching.Spikes`

### Electrode denormalization on `GlobalUnit`
- Moved peak electrode from `UnitMatching.Unit` to `GlobalUnit` itself
- FK target: `ephys.ProbeType.Electrode` (not `ElectrodeConfig.Electrode`) — the electrode is a property of the probe geometry, not the recording config
- Updated each time a new session is matched (latest session = best estimate)
- This makes the **unit roster query trivial**: `GlobalUnit & insertion_key` → one row per unit with electrode, no joins

### Query UX design
The two primary user-facing tables are `GlobalUnit` (unit roster) and `UnitMatching.Spikes` (spike data). The typical workflow:

```python
insertion_key = {"experiment_name": "exp02", "subject": "BAA-1104548", "insertion_number": 1}

# 1. Unit roster — one row per unit, with electrode
GlobalUnit & insertion_key

# 2. Spike times — one row per chunk (unique index enforced)
(UnitMatching.Spikes & insertion_key & {"global_unit": 5}).fetch(
    "chunk_start", "spike_times", "spike_count", order_by="chunk_start"
)

# 3. Filter by chunk range — for alignment with other data streams
(UnitMatching.Spikes & insertion_key & {"global_unit": 5}
 & f'chunk_start BETWEEN "{start}" AND "{end}"'
).fetch("chunk_start", "spike_times", order_by="chunk_start")
```

`UnitMatching.Unit` (session→unit linkage) is pipeline plumbing — rarely queried directly by end users.

### Temporal ordering enforcement
- `key_source` gates to earliest unprocessed `block_start` per insertion via `MIN(block_start)` aggregation
- `make()` guard as defense-in-depth against direct `make(key)` calls
- Safe under `populate(reserve_jobs=True)` — different insertions parallelize, same insertion serializes

Full spec: [`SPEC_unit_matching.md`](https://github.com/SainsburyWellcomeCentre/aeon_mecha/blob/tn/ephys_revamp_v2/aeon/dj_pipeline/docs/specs/SPEC_unit_matching.md)

esutlie (approved):
Looks great! If you take it out of draft mode, I will merge it in!

ttngu207:
> Looks great! If you take it out of draft mode, I will merge it in!

Done

ttngu207:
@esutlie  — heads up on a design decision I'm settling on after going back and forth on this.

**I'm restoring `subject` to the `ProbeInsertion` PK** (reverting the v2 removal), and redesigning `EphysEpoch` and `EphysChunk` accordingly. The full rationale is in the updated spec (`SPEC_unit_matching.md`), but the core reasoning is:

### 1. Respecting the physical experimental setup

A `ProbeInsertion` is a **surgical event** — a specific probe implanted into a specific subject's brain. Subject is a natural part of that identity, not an afterthought to be linked separately. The current `EphysSubject` (Manual, no enforcement) approach can silently break if left unfilled, and critically, **it cannot represent probe swaps** — if the same physical probe is moved from Subject1 to Subject2, the no-subject design merges two different neuron populations into one insertion. With subject in PK, these are naturally two distinct insertions with separate GlobalUnit lineages.

### 2. More natural queries for downstream users

Every neuroscience analysis correlates spikes with behavior, which requires subject identity. With subject in the PK chain, every table from `EphysChunk` down to `GlobalUnit` and `UnitMatching.Spikes` carries subject directly:

```python
# Direct — no extra join needed
GlobalUnit & {"subject": "BAA-1104545"}
UnitMatching.Spikes & {"subject": "BAA-1104545", "global_unit": 5}
```

vs. the current design where every subject-based query requires routing through `EphysSubject`:

```python
# Current — extra join on every query
UnitMatching.Spikes & (EphysSubject & {"subject": "BAA-1104545"})
```

### What changes

- `ProbeInsertion`: `subject` back in PK, `probe_label` moves to `EphysEpoch.Insertion`
- `EphysSubject`: eliminated
- `EphysEpoch`: redesigned as a per-epoch probe registry (reads subject-probe mapping from a per-epoch file or carries forward from previous epoch; also auto-creates `Probe` entries since probe_type is now discoverable from files)
- `EphysChunk`: gains a FK to `EphysEpoch` for epoch traceability
- `ingest_chunks()`: epoch-aware with a guard — skips files without subject-probe mapping

The operational requirement is that subject-probe association must be provided (via per-epoch file or manual insert) before ingestion. This is reasonable — probe surgery is a planned procedure, and the experimenter always knows which subject has which probe.

See commit `15d7a33` for the full updated spec.

ttngu207:
## Design Discussion: Sequential vs. Pair-Based Unit Matching

Both designs are now implemented side-by-side in `spike_sorting.py`. This comment captures the analysis to help the team decide which to keep.

### The Two Designs

**Sequential (`UnitMatching`)**: A Computed table whose `key_source` auto-selects the next unprocessed block in temporal order per (insertion, paramset). `make()` compares the new block against ALL overlapping previously-matched blocks. Users just call `populate()`.

**Pair-based (`UnitMatchingPair` + `PairMatching`)**: A Manual table where users explicitly specify which two blocks to compare (+ paramset). A downstream Computed table performs the matching. At least one block must be already matched (anchor), unless it's the seed pair.

### End-User Experience: Common Case (100 blocks)

| Step | Sequential | Pair-based |
|------|-----------|------------|
| Setup | Nothing | ~99 manual `UnitMatchingPair.insert1()` calls |
| Run | `UnitMatching.populate()` — one call | `PairMatching.populate()` — but most pairs fail on first pass because the anchor isn't matched yet. Requires multiple rounds of `populate()` or manual `make()` calls in the right order |
| New blocks arrive | `populate()` again | User must create new pairs manually |
| Try different parameters | New paramset → `populate()` | New paramset → 99 new manual inserts |

### The Motivating Use Case for Pair-Based

> "I spike sort and curate 10 blocks (0–9), then want matching to start from the middle — (4,5) as seed — and propagate outward: 4→3→2→1→0 and 5→6→7→8→9."

This is compelling: the scientist picks the seed block based on domain knowledge (best signal quality, most stable recording period), not arbitrary temporal order. If block 0 has poor signal, anchoring everything to it produces worse global unit assignments.

**The sequential design can't do this** — its `key_source` always picks `MIN(block_start)`, forcing block 0 first.

**The pair-based design can express this** — create (4,5) as seed, then (3,4), (5,6), (2,3), (6,7), etc. But `populate()` still can't process them in the right order without multiple rounds or manual `make()` calls.

### A Possible Middle Ground

The seed/direction problem could also be solved by extending the sequential design:

- Add a `seed_block_start` (or `propagation_strategy`) to `UnitMatchingParamSet` or a small config table
- Modify `key_source` to start from the seed and propagate outward (e.g., by distance from seed) instead of always picking `MIN(block_start)`
- Same zero-manual-work UX: `populate()` just works, but from the user-chosen starting point

This gives scientists the control they want ("start from block 4") without the manual burden of wiring every pair.

### Tradeoff Summary

| Aspect | Sequential | Pair-based |
|--------|-----------|------------|
| Common case UX | One `populate()` call | N-1 manual inserts + multiple `populate()` rounds |
| Seed block choice | Not currently supported (always `MIN(block_start)`) — but could be added | Fully supported |
| Bidirectional propagation | Not currently supported — but could be added via `key_source` | Supported (user creates pairs in desired order) |
| Non-adjacent matching | Already handled (compares ALL overlapping blocks, not just adjacent) | Also supported |
| Correctness guarantee | Enforced by `key_source` + `make()` guard | Relies on user creating pairs in valid order |
| Re-curation recovery | Re-process from changed block forward (heavy-handed but correct) | Can surgically re-process specific pairs (but user must identify affected pairs) |
| Parameter exploration | New paramset → `populate()` | New paramset → N-1 new manual inserts |
| `populate()` behavior | Processes chain in one pass | Fails for most pairs until chain is built; needs multiple rounds |

### Recommendation for Discussion

The pair-based design's real contribution is the insight that **scientists need control over the seed block and propagation direction**. But the mechanism (manually specifying every pair) is more burden than necessary.

Consider: **keep the sequential design and extend it with seed/direction control.** This gives scientists the meaningful control (where to start, which direction) without the operational cost of managing individual pairs. The `key_source` modification is straightforward and the UX remains `populate()`.

The pair-based code (`UnitMatchingPair` + `PairMatching`) is committed for reference and can be kept or removed after team discussion.

ttngu207:
## Decision: Seed-based UnitMatching, PairMatching removed

@esutlie please take a look

After discussion, we rejected the pair-based `PairMatching` alternative in favor of enhancing `UnitMatching` with **seed-based bidirectional propagation**.

### What changed (`8f3cff0`)

**`UnitMatchingParamSet`** — added `seed_block_start: datetime(6)`. Scientists pick the block with best signal quality as the seed. Different seeds → different paramset entries.

**`key_source`** — rewritten as pure SQL (DataJoint query expressions). Yields at most 3 keys per (insertion, paramset):
- The seed block (if unprocessed)
- Forward frontier: `MIN(unprocessed block_start)` after the processed region
- Backward frontier: `MAX(unprocessed block_start)` before the processed region

**`make()` guard** — replaced strict temporal ordering with:
- First block must equal `seed_block_start`
- Subsequent blocks must overlap with at least one already-processed block

**Removed** — `UnitMatchingPair` (Manual) and `PairMatching` (Computed + Unit/Spikes Parts) — the entire pair-based alternative (~430 lines).

**Spec** — updated to reflect seed-based propagation, removed Alternative Design section, added design decision rationale.

### Why seed over pairs

The pair-based design was functionally equivalent to sequential matching but added schema complexity (projected nullable FKs, anchor determination logic, manual pair insertion). The seed-based approach gives the same user control (choosing where to start) with automatic propagation and a simpler schema.

aeon/dj_pipeline/docs/specs/SPEC_unit_matching.md
```diff
@@ -0,0 +1,699 @@
...
+
+When two ephys blocks overlap in time, the same `EphysChunk` (1-hour raw data segment) is sorted independently by both blocks. For a global unit present in both blocks, `SyncedSpikes.Unit` contains spike times from both blocks for the overlapping chunks. Without a convention, `Spikes` would contain duplicate rows for the same `(global_unit, chunk)` from different blocks.
+
+Duplicates cause incorrect downstream analysis: inflated spike counts, artificial short ISIs, and wrong firing rate estimates.
+
+### Convention: Earlier Block Owns
+
+**Rule: for each `(global_unit, chunk)` pair, the first block (in processing order) to write a `Spikes` row wins. Later blocks skip that pair.**
+
+Since blocks are processed in temporal order, "first to process" = "earlier block" for chunks in the overlap region.
```
esutlie: @ttngu207 No longer true. We can just remove this line.

```suggestion
```

esutlie (approved):
Looks good to me! Just one small comment on updating the spec.

PR #553, opened 2026-04-27
Forward-compatible MariaDB json column workaround

esutlie:
## Summary

MariaDB aliases `json` columns to `longtext`, which causes DataJoint to skip automatic JSON serialization on insert and deserialization on fetch. This adds a runtime workaround that checks `attr.json` at populate time and serializes manually when needed. It works on both MariaDB and MySQL, and will stop serializing automatically once DataJoint fixes the root cause.

Tested on HPC against aeon-db (MariaDB 10.3). All 9 stream tables populate successfully and codec fetch returns a DataFrame as expected.

See datajoint/datajoint-python#1438

## Changes

- **New: `aeon/dj_pipeline/utils/json_serialization.py`** - `json_serialized(value, needs_serialize)` helper that serializes to JSON only when DataJoint won't
- **Modified: `aeon/dj_pipeline/utils/codec.py`** - `encode()` now accepts both dict (MySQL) and pre-serialized string (MariaDB). `decode()` handles both string and dict input on fetch.
- **Modified: `aeon/dj_pipeline/utils/streams_maker.py`** - `make()` template checks `attr.json` once per call and wraps all json column values with `json_serialized()`

### DIFF
--- a/aeon/dj_pipeline/utils/codec.py
+++ b/aeon/dj_pipeline/utils/codec.py
@@ -7,6 +7,8 @@ Individual data columns store JSON summary stats (min, max, mean, dtype, count).
 The `stream_df` column uses <aeon_stream> codec for lazy DataFrame loading.
 """
 
+import json
+
 import datajoint as dj
 import numpy as np
 import pandas as pd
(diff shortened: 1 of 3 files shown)


AdrianRoggenbach (approved):
Looks good to me, I will test these changes in the foragingABC_analysis repo, also with new experimental data.

PR #562, opened 2026-05-11
ONIX IMU pipeline + ephys SyncModel refactor

ttngu207:
## Summary

Promotes the per-chunk HARP↔ONIX sync regression from a Part of `EphysChunk` (with `<attach>` model + ONIX bounds) to a first-class `EphysSyncModel(dj.Manual)` table keyed by `(EphysEpoch, sync_start)`. Refactors `EphysChunk.ingest_chunks` to query SyncModel rows instead of fitting regressions in-memory. Adds a new `OnixImuChunk(dj.Imported)` table that surfaces the four Bno055 IMU streams via a structural-only `<aeon_onix_stream>` codec, with HARP-indexed access exposed via an explicit `OnixImuChunk.synced_df(key)` classmethod.

Full design rationale: [`docs/specs/SPEC_ONIX_IMU_PIPELINE.md`](docs/specs/SPEC_ONIX_IMU_PIPELINE.md).

## Architecture

```
EphysEpoch
   │
   ▼
EphysSyncModel  (Manual + ingest classmethod, re-runnable)
   │
   ├─► EphysChunk   (Manual + ingest_chunks classmethod, re-runnable)
   │     ├─► .File       (AmplifierData + Clock files)
   │     └─► .SyncModel  (Part — link-only FK to EphysSyncModel)
   │
   └─► OnixImuChunk (Imported, default key_source = EphysSyncModel)
         — single row per sync window; 4 Bno055 streams merged on sample index
```

- `EphysSyncModel` PK uses **observed** `harp_time[0]` from the CSV (stable across re-runs, master-clock domain).
- `OnixImuChunk.stream_df` codec returns **ONIX-clock-indexed** DataFrame (structural only — no regression). `OnixImuChunk.synced_df(key)` is the opt-in HARP-indexed path.
- Each `OnixImuChunk` row has 13 JSON summary attrs (one per IMU column: euler/gravity_vector/linear_acceleration/quaternion components) plus `sample_count`, `timestamps`, and the codec column — matches the auto-stream-table pattern in `streams_maker`.

## Test plan

- [x] Unit suite: 74 passed (`uv run pytest -m unit tests/`)
- [x] Integration suite: 68 passed, 3 skipped (golden-dataset gated) (`uv run pytest -m integration tests/dj_pipeline/`)
[...]

PR #563, opened 2026-05-13
Add ephys pipeline user guide and fix 10 bugs found during HPC testing

esutlie:
## Summary

- Add a complete end-to-end ephys pipeline user guide (`docs/ephys_guide/`) with 7 step files and a wrapper notebook, validated against Adrian's golden baseline dataset (abcGolden01)
- Implement `read_probe_assignments()` for JSON-based probe-to-brain-region mapping with carry-forward logic and override support for read-only filesystems
- Add `"raw-ephys"` directory type so experiments with separate behavior/ephys rigs (e.g., AEON3 + AEONX1) can register both directories independently
- Fix 10 bugs discovered when running the pipeline against real Neuropixels data for the first time — documented in `docs/ephys_guide/bugs_found.md`
- Update `spike_sorting.py` for DJ 2.x compatibility (codec syntax, core types, MariaDB workaround)
- Update 6 environment setup docs to replace conda/pip references with uv

## Guide structure

| File | Description |
|------|-------------|
| `step00_setup.md` | HPC access, environment, database config |
| `step01_register_experiment.py` | Register experiment, directories, epochs, sync models, chunks |
| `step02_define_blocks.py` | Define ephys blocks (20h max, 5h overlap) |
| `step03_spike_sorting.py` | Preprocessing (CPU) + Kilosort4 sorting (GPU via SLURM) |
| `step04_curation.md` | Manual curation with SpikeInterface GUI |
| `step05_unit_matching.py` | Cross-block unit matching |
| `step06_analysis_examples.py` | Spike times, behavioral events, PETH histograms |
| `ephys_guide.ipynb` | Wrapper notebook linking all steps |
| `bugs_found.md` | All 10 bugs found during HPC testing, with root causes and fixes |

### New: `"raw-ephys"` directory type

Experiments like `abcGolden01` have behavior data on one rig (AEON3) and ephys data on another (AEONX1). `Epoch.ingest_epochs()` now auto-discovers both:
- `"raw"` directories: scans for CameraTop CSV chunks (behavior)
- `"raw-ephys"` directories: enumerates epoch directories containing NeuropixelsV2 subdirectories (ephys)

### New: `read_probe_assignments()`

Reads `probe_assignments.json` from the epoch directory to map probe serial numbers to brain regions. Supports carry-forward (reuse assignments from previous epochs in the same experiment) and an `override_dir` kwarg for read-only filesystems like Ceph.

### Bug fixes (details in `bugs_found.md`)

**Metadata/data issues (bugs 1-5):**
- `get_probe_id()` — wrong metadata path, no disabled probe detection, Windows path parsing on Linux
- `HarpSyncModel.Reader` — stored HARP seconds (~3.86B) as ONIX timestamps instead of actual ONIX clock values
- Guide script missing sync model ingestion step

**DJ 2.x compatibility (bugs 6-10):**
- `spike_sorting.py` — `filepath@dj_store` → `<filepath@dj_store>`, `blob@dj_store` → `<blob@dj_store>`
- `spike_sorting.py` — `float` → `float64`, `int` → `int32` (core types)
- `spike_sorting.py` — `JSON` → `json` (case-sensitive type map)
- `ephys.py` — `EphysChunk.File` stored `directory_type: "raw"` instead of `"raw-ephys"`
- `spike_sorting.py` — `<filepath@dj_store>` codec returns dict for JSON column, MariaDB 10.3.28 can't serialize it. Workaround: try/except on `File.insert()`, fallback path reconstruction on `File.fetch()`. Forward-compatible with MySQL upgrade.

## HPC validation status

- **Step 1 (register experiment):** Passed — 12 chunks, 3 sync models, 2 epochs
- **Step 2 (define blocks):** Passed — 3 blocks (30min/10min overlap)
- **Step 3 (preprocessing):** In progress — block 1 done, blocks 2-3 writing
- **Steps 4-6:** Not yet tested

## Depends on

- #562 (ONIX IMU pipeline + ephys SyncModel refactor) — this branch is built on top of `tn/onix-imu-pipeline`


ttngu207:
Thanks for the real-data pass — bugs 1-4 and 9 are exactly the kind of thing synthetic tests can't catch, and bug 4 specifically corrects a mistake of mine on `tn/onix-imu-pipeline`. The `data.index.values` "fix" I made in `1436fbb` was wrong: I misread the `Csv.read()` contract (first column → index via `index_col=0`, `names=` covers the remaining columns), and my 3-column synthetic fixture hid it. Your revert is the right call.

A few things to address before merge:

**Critical**
1. `sorting_file` fallback path in `PostProcessing.make_fetch` (`spike_sorting.py:477`) points to `output_dir / "si_sorting.pkl"` but `SpikeSorting.make_compute` writes to `output_dir / "spike_sorting" / "si_sorting.pkl"`. When the MariaDB workaround skips the insert, this fallback `FileNotFoundError`s — defeats the workaround.
2. `OnixImuChunk.make` and `OnixStreamCodec.decode` still hardcode `directory_type="raw"`. For AEON3+AEONX1 experiments the IMU pipeline will silently insert `sample_count=0` rows. Either thread `directory_type` through or fall back to `"raw-ephys"` when the device dir isn't under `raw`.
3. My `test_ephys_reader_unit.py` fixture is 3-column — 2 tests fail under your reader fix. I'll send a fixture update on `tn/onix-imu-pipeline` so CI is green when you rebase.
4. `Epoch.ingest_epochs` now requires both `raw` and `raw-ephys` directories to be registered. For experiments where ephys data lives in the same place as behavior, the two rows can point to the same path; for split rigs (AEON3+AEONX1) they differ. Worth documenting in the `Experiment.Directory` / `ingest_epochs` docstrings and flagging in the PR description — existing AEONX1-only setups (currently registered as `raw`) need to add a matching `raw-ephys` row pointing at the same path or new epochs will silently stop being ingested.

**Should fix**
- Carry-forward `KeyError` when a new probe appears in a later epoch — `read_probe_assignments` should raise informatively.
- `except Exception:` in `make_fetch` fallbacks → narrow to `dj.errors.DataJointError`.
- Fork-crash fix only applied to `write_binary_recording`. `PostProcessing.make_compute` and `SIExport.make` still spawn workers via guide-configured `n_jobs=0.8`. Either apply the same `n_jobs=1` or document the workaround for users.
- Direct tests for `get_probe_id` (valid V2, disabled, Windows path, V2Beta, missing `Devices` key) and `read_probe_assignments` (override_dir, carry-forward, missing-file) — those bug paths are currently untested.

**Minor / nits**
- `ephys_utils.py:182` comment says "Priority 2: Carry-forward" — should be Priority 3.
- `ephys_utils.py:187` `fetch(as_dict=True, order_by=...)` → `.to_dicts(order_by=...)` for DJ 2.x consistency.
- `ephys.py:193` unused loop var `label`.
- `scripts/ephys_v2_validate.py:103` still only accepts `directory_type == "raw"` — add `"raw-ephys"`.
- `probe_assignments.json` version check is presence-only; tighten to `== 1` to leave room for v2.

ttngu207:
Follow-up on critical points 2 and 4 — talking it through, I think there's a simpler model that handles both cleanly:

**Policy:**
- `raw` is mandatory for every experiment.
- `raw-ephys` is **optional**; register it when ephys data lives somewhere other than `raw`, or when you want to be explicit that ephys exists. Combined-rig experiments (ephys in the same dir as behavior) don't need to register it.
- Downstream ephys resolution: prefer `raw-ephys` if registered, fall back to `raw`. If neither has an ephys device subdir, no ephys data.
- Invariant: every ephys epoch has a corresponding behavior epoch (CameraTop CSV exists). No orphan ephys epochs.

**What this lets us simplify:**
- **Drop pass 2 in `Epoch.ingest_epochs`** (~40 lines). Under the no-orphan-ephys invariant, pass 1 catches every epoch via CameraTop. Pass 2 is gratuitous, and removing it eliminates the precedence/idempotency edge cases between the two passes.
- **Replace hardcoded `directory_type="raw"`** in `EphysSyncModel.ingest`, `EphysChunk.ingest_chunks`, `OnixImuChunk.make`, `OnixStreamCodec.decode` with a single resolver — e.g. `resolve_ephys_dir(key)` that tries `raw-ephys` first, falls back to `raw`. That folds critical points 2 (OnixImuChunk silently broken on split-dir) and bug 9 (`EphysChunk.File` directory_type) into one helper instead of threading the type through every call site.
- **`EphysChunk.File.directory_type`** should then store whatever the helper resolved (`raw-ephys` for split-rig, `raw` for combined-rig) rather than a hardcoded value.

Optional: when ephys binaries exist on disk for an epoch that has no `Epoch` row (invariant violated), log a warning in `EphysSyncModel.ingest` to surface the data-integrity issue rather than silently ingesting nothing.

Does this match your intent? If yes, it'd shrink the PR a fair bit and clean up the directory-type story.

ttngu207:
Closing, replaced by #564

PR #564, opened 2026-05-14
Add ephys pipeline runbooks

esutlie:
## Summary

- Complete ephys pipeline implementation: epoch ingestion, block definition, spike sorting (Kilosort4 via SLURM), post-processing, curation, unit matching, and analysis examples
- End-to-end user guide (`docs/ephys_guide/`) with runnable scripts (steps 00-06) and a wrapper notebook
- Split raw directory support: `raw-ephys` directory type for rigs that record behavior and ephys on separate machines (e.g., AEON3 + AEONX1)
- `read_probe_assignments()` with JSON-based probe configuration and carry-forward across epochs
- DJ 2.x compatibility fixes throughout (type syntax, codec, filepath handling)
- MariaDB 10.3.28 compatibility for filepath/blob codecs
- SLURM job array scripts for GPU spike sorting
- Onix IMU pipeline and stream codec (from Thinh's `tn/onix-imu-pipeline` work)
- Tests for probe assignments, ephys readers, IMU pipeline, and codec

## Validation

- Steps 1-3 validated end-to-end on HPC with the abcGolden01 golden dataset (12 sorting tasks completed across 3 blocks x 4 shanks)
- Step 4 (post-processing) in progress on HPC
- Steps 5-6 (unit matching, analysis) not yet validated — expected in follow-up PR

Replaces #562. Supersedes draft #563.


ttngu207:
Two from my earlier review that I think are still open:

**C2 — `OnixStreamCodec.decode` still hardcodes `directory_type="raw"`** (`codec.py:155`). Under the new policy where `raw-ephys` is the source of truth for ephys data (which `EphysSyncModel.ingest`, `EphysChunk.ingest_chunks`, and `OnixImuChunk.make` all use via `resolve_raw_dir_and_epochs`), the codec is the odd one out. For a split-rig experiment (AEON3+AEONX1), `OnixImuChunk.make` correctly finds the device dir under `raw-ephys` and inserts a populated row, then any fetch of `stream_df` re-resolves the device dir via the codec — which looks under `raw`, doesn't find it, and silently returns an empty DataFrame. Suggest the codec also look up `raw-ephys` (or call `resolve_raw_dir_and_epochs` directly) so the policy is consistent end-to-end.

**C3 — `tests/schema/test_ephys_reader_unit.py` fixture is still 3-column.** The `HarpSyncModel.Reader` fix in `dc4188f` is correct (Aeon CSVs have 4 columns: Aeon timestamp as the unnamed first column, then `clock`/`hub_clock`/`harp_time`), but the test fixture I wrote earlier on `tn/onix-imu-pipeline` only writes 3 columns. Running `uv run pytest -m unit tests/` returns `78 passed, 2 failed` on this branch. Easy fixture update — happy to send a commit since it's my test.

ttngu207:
Mild suggestion — rename `docs/ephys_guide/` → `docs/ephys_runbooks/`. These are step-by-step executable scripts users run against their own data, not a tutorial / explainer doc. "Runbook" reads more accurately to me than "guide" for what they actually are. Totally non-blocking; flag for when/if you're already touching those files.

lochhh (commented):
## Code Review (from Claude)

Big PR — impressive scope. Solid test coverage, the `bugs_found.md` transparency is genuinely useful, and the `probe_assignments_override_dir` pattern for read-only Ceph is clean. Steps 1–3 validated on HPC is good signal. Steps 5–6 not yet validated; tagging those as known-incomplete is the right call.

Issues below, ordered by severity.

---

### CRITICAL

**A. `PostProcessing.make_compute()` bare `return` on no-units path — `aeon/dj_pipeline/spike_sorting.py:529`**

When `si_sorting.unit_ids.size == 0`, the method hits `return` without a value. The return annotation is `Tuple[Path, datetime, float]`, and `make_insert()` at line 556 unpacks three positional args. When DJ calls `make_insert(key, *result)` and `result is None`, this crashes with `TypeError: cannot unpack non-iterable NoneType object`.

Fix:
```python
if not has_units:
    logger.info("No units found. Skipping sorting analyzer.")
    analyzer_output_dir.mkdir(parents=True, exist_ok=True)
    return analyzer_output_dir, execution_time, 0.0
```

**B. Hardcoded analysis repo path in `run_aeon_spike_sorting.sh:48`**

```bash
cd ~/ProjectAeon/foragingABC_analysis
```

This breaks for every user other than the original author. Fix options:
- `cd "$SLURM_SUBMIT_DIR"` (use submit directory)
- Require `ANALYSIS_REPO` env var: `cd "${ANALYSIS_REPO:?Set ANALYSIS_REPO to your analysis repo path}"`
- At minimum: uppercase `# EDIT THIS LINE` comment so it's unmissable

---

### HIGH

**C. Array mutation in `indices2syncedtimes()` — `spike_sorting.py:1066`**

```python
spk_ind -= spk_ind[0]  # in-place, modifies original if spk_ind is a view
```

Lines 1058 and 1060 create `spk_ind` via boolean indexing, which in NumPy returns a copy — so this is safe today. But it's fragile; the next refactor that changes the slicing pattern could introduce data corruption silently. Add `.copy()` to make the intent explicit:

```python
spk_ind = spk_ind.copy()
spk_ind -= spk_ind[0]
```

**D. `clear_job()` clears ALL errored jobs for the table, not just the target key — `scripts/run_aeon_spike_sorting.py:133`**

In a SLURM job array, multiple tasks may fail concurrently. Task 1 resubmitting with `CLEAR_JOB=True` silently wipes Task 5's error state, making it invisible for recovery. The safest fix is to delete only the row matching the current key's status entry, or document prominently that this is intentionally a table-wide clear.

**E. Bare `except Exception` in `EphysChunk.ingest_chunks()` — `ephys.py:532,576`**

Two broad catches here:
- Line 532: catching any error in HARP time resolution — could hide `joblib.load` failures, corrupt model files, or numpy errors silently
- Line 576: catching any error in the 3-table transaction — could hide DB constraint violations

Both log and continue, which is correct behaviour, but the catch should be narrowed to expected failure modes. Suggest:
```python
except (FileNotFoundError, OSError, ValueError, RuntimeError) as e:  # line 532
except (dj.errors.DataJointError, OSError, TypeError) as e:  # line 576
```

---

### MEDIUM

**F. MariaDB `except TypeError` may swallow unrelated errors — `spike_sorting.py:306,439,572,634`**

The four `try/except TypeError` blocks on `File.insert()` are correct for the MariaDB JSON codec issue, but a bare `TypeError` also matches wrong-type field values, bad list comprehension output, etc. Narrowing the guard reduces the silent-failure surface:

```python
except TypeError as e:
    if "codec" not in str(e) and "JSON" not in str(e):
        raise
    logger.warning("...File insert skipped (MariaDB JSON compat)")
```

**G. Ephys epoch insert has no transaction — `acquisition.py:335–342`**

Behavior epochs (line 264) are wrapped in `with cls.connection.transaction:`. Ephys epochs at line 335 are not. A failure mid-insert leaves an orphan `Epoch` master row without part rows.

**H. Hardcoded `/ceph/aeon/` root in guide step01 — `docs/ephys_guide/step01_register_experiment.py:119`**

```python
ceph_root = "/ceph/aeon/"
```

Used to compute `directory_path` stored in the DB. If Ceph is mounted at a different prefix, every stored path is wrong and downstream queries will fail to resolve files. Should be derived from the configured repository path or at minimum flagged as "MUST MATCH YOUR CEPH MOUNT".

**I. `n_jobs=1` hardcode is undocumented — `spike_sorting.py:267,605`**

`write_binary_recording` uses `n_jobs=1` but PostProcessing at line 512 uses `n_jobs=-1`. The asymmetry is correct (fork+BLAS crash documented in `bugs_found.md` #11) but there's no comment in the code. Future readers will "fix" this to improve performance and reintroduce the crash. Add:

```python
job_kwargs = {"n_jobs": 1, "chunk_duration": "2s"}  # n_jobs=1: fork+BLAS crash on SWC HPC (bugs_found.md #11)
```

---

### LOW / NITS

**J. `_chunk_total_duration` executes a DB query but is never used — `ephys.py:661–667`**

The comment says "reserved for future validation" but the variable is dead code behind a DB round-trip. Either implement the validation or remove the block.

**K. `locate_bno055_chunk_index()` uses exact uint64 match — `utils/onix_imu.py:108`**

```python
if int(first_clock_arr[0]) == int(onix_ts_start):
```

Silently returns `None` if timestamps have any jitter or rounding. Document the exact-match requirement in the docstring so future callers know this isn't fuzzy matching.

**L. SLURM resource spec is likely oversized for short blocks — `run_aeon_spike_sorting.sh:21–22`**

`--mem=256G` and `--time=7-08:00:00` will queue much longer than needed for 30-minute test blocks. Consider adding a comment with typical values (e.g. `# 48G and 4h for <1hr blocks`) to help users tune their own submissions.

**M. `slurm_output/` mkdir happens after SBATCH output directive — `run_aeon_spike_sorting.sh:24,40`**

SBATCH tries to open the output file before the script body runs. If `slurm_output/` doesn't exist the job silently loses its output. Either commit the directory to the repo (with a `.gitkeep`) or document that it must be pre-created.

---

### Positive callouts

- Test coverage is strong: unit tests, integration tests, synthetic fixture helpers — `_synthetic_ephys_fixtures.py` is well-structured
- `read_probe_assignments()` carry-forward logic is clean and safe
- `OnixStreamCodec`: stream-group guard order is correct (rejects unsupported groups before Bno055 lookup)
- Transaction wrapping in `EphysChunk.ingest_chunks()` for the 3-table insert is the right pattern
- `probe_assignments_override_dir` fallback for read-only Ceph is elegant
- `PureWindowsPath` for probe serial extraction is the correct cross-platform approach for Windows-format metadata paths
- `bugs_found.md` — rare and valuable transparency; keep this pattern

tests/dj_pipeline/utils/test_onix_codec_integration.py
lochhh, line 1: These tests can be merged into test_codec_integration.py

tests/fixtures/ephys/ephys_factories.py
lochhh, line 1: consider moving this into tests/fixtures/ephys for better organisation and maybe even rename this to ephys_factories to make it clear these are factory functions not fixtures? the functions can also be public since pytest will only collect test_* functions

esutlie:
@lochhh Thanks for the thorough review! Here are my thoughts:

### Fixing

**A (PostProcessing bare return)** — The bare `return` means `make_insert` is never reached, so the key silently re-runs on every `populate()`. I'll fix this so the session key flows through the whole chain even when there are no units — PostProcessing and SIExport will both get zero-units guards, matching what SortedSpikes already does.

**B (hardcoded SLURM path)** — Agreed, and the bigger issue is that the current flow asks users to edit tracked files in the submodule, which dirties their git state. What do you think about changing `print_slurm_config()` to a `write_slurm_config()` that writes both `.py` and `.sh` directly into the user's working directory, pre-filled from the database? The `.sh` script would use `$SLURM_SUBMIT_DIR` instead of a hardcoded path, so users just run `sbatch` without editing anything.

**D (clear_job scope)** — Will add `& key` to the restriction.

**H (hardcoded Ceph root)** — Will read from `get_repository_path("ceph_aeon")` instead of hardcoding.

**I (n_jobs=1)** — I'll add a comment for now and investigate the error logs to see if there's a way to let SpikeInterface do multipool processing on the HPC.

**J (_chunk_total_duration)** — I'll remove it.

**L (SLURM resources)** — I'll add a comment with typical values.

**M (slurm_output dir)** — I'll add `.gitkeep`.

### Leaving as-is

**C (array mutation)** — Boolean indexing in NumPy always returns a copy, so `spk_ind` is safe to modify in place.

**E (bare except Exception)** — These log-and-skip instead of crashing, which I think is the preferred behavior.

**F (MariaDB TypeError)** — This needs a more robust fix for similar inserts in the tables post spike-sorting, so these `TypeError` exceptions will be reverted in a future commit.

**G (ephys epoch transaction)** — This is a single `insert1()` which is already atomic, so we don't need a transaction. The behavior epochs at line 264 need one because there are two operations that need to be paired in one transaction.

**K (exact uint64 match)** — ONIX timestamps are exact hardware clock counters with no jitter, so exact match comparison is correct here.

esutlie:
Hey all, I finished testing the full pipeline end-to-end on the HPC. I've applied the edits from your reviews and squashed a few more bugs that came up along the way (details in `docs/ephys_runbooks/bugs_found.md`).

One thing to note, there's a temporary monkey-patch in `aeon/dj_pipeline/__init__.py` (lines 34-40) for a MariaDB compatibility issue. DJ 2.x's external store codecs (`<filepath@store>`, `<blob@store>`) return a Python dict on insert, and MariaDB doesn't know how to serialize it. The patch just teaches pymysql to serialize dicts as JSON strings before they hit the database. It's inert on MySQL so it won't cause any issues down the line. It will simply become dead code when we upgrade to MySQL.

Also, let me know what you think about the `write_slurm_config()` idea from my earlier comment. The SLURM templates work as-is but they have hardcoded paths that users need to edit. This would just make the step 3 runbook autogenerate the scripts with the right keys based on a template. Happy to do that as a follow-up PR if it sounds useful.

lochhh:
> Also, let me know what you think about the `write_slurm_config()` idea from my earlier comment. The SLURM templates work as-is but they have hardcoded paths that users need to edit. This would just make the step 3 runbook autogenerate the scripts with the right keys based on a template. Happy to do that as a follow-up PR if it sounds useful.

Thanks @esutlie ! 
I like your idea and think we should replace `print_slurm_config()` with `write_spike_sorting_scripts(experiment_name, subject, outdir=None)` to auto-generate the `run_aeon_spike_sorting` python and slurm scripts. In particular the function should:

- Query SortingTask (same as current)
- Write `run_aeon_spike_sorting.py` to outdir (default: CWD): pre-filled `_base` and `keys` from DB, boilerplate generated as string
- Write `run_aeon_spike_sorting.sh` to outdir, generated as string with:
  - #SBATCH --array=1-{n_keys} auto-set from query result
  - cd "$SLURM_SUBMIT_DIR" replacing hardcoded analysis repo path
  - SCRIPT_PATH pointing to $SLURM_SUBMIT_DIR/run_aeon_spike_sorting.py
- Print confirmation, e.g. 
  ```
  Wrote: /path/to/run_aeon_spike_sorting.py (12 keys)
  Wrote: /path/to/run_aeon_spike_sorting.sh

  Submit with:
      sbatch run_aeon_spike_sorting.sh
  ```

Both of these files land in the user's analysis repo (outside the submodule) so that we don't end up with dirty git state.

With the above in place, we can either
- remove both .sh and .py from aeon_mecha, or
- (preferably) move `run_aeon_spike_sorting.sh` from repo root to `docs/ephys_runbooks/` (with $SLURM_SUBMIT_DIR already substituted, showing the generated form) and move `aeon/dj_pipeline/scripts/run_aeon_spike_sorting.py` to `docs/ephys_runbooks/` (keep golden dataset values: concrete, matches the runbook throughout). In `step03_spike_sorting.py` the docstring points to them, e.g. "see docs/ephys_runbooks/run_aeon_spike_sorting.{py,sh} for example output"

ttngu207:
Regarding the SLURM refactoring/optimization above, I'd suggest for that to be a separate, follow up PR.

ttngu207 (dismissed):

lochhh (dismissed):

pyproject.toml
esutlie, line 193: @ttngu207 @lochhh This is the line that is causing the CI tests to try importing `aeon_exp_foragingABC`. This was introduced before this PR, so I'm not sure what changed.

lochhh: Fixed: [Update uv.lock](https://github.com/SainsburyWellcomeCentre/aeon_mecha/pull/564/commits/374b7140433fc10bf535e05fb1c528ea1cd6abe0)

lochhh (approved):

PR #572, opened 2026-05-15
Mock DJ for all tests marked "unit"

lochhh:
The current setup will only mock DataJoint when the tests are run with `-m unit`. 

Running `uv run pytest tests/dj_pipeline/utils/test_stats_unit.py` for instance will result in the following error:
> FAILED tests/dj_pipeline/utils/test_stats_unit.py::TestColumnStats::test_column_stats[numeric array] - pymysql.err.OperationalError: (2003, "Can't connect to MySQL server on 'localhost' ([WinError 10061] No connection could be made because the targ...

Instead of checking for the `-m` marker, this PR adds the `mock_dj_for_unit` fixture that auto-runs on every test. 
For non-unit tests: no-op, yield, done. 
For unit tests:
1. save datajoint and aeon.dj_pipeline from sys.modules (module cache) before removing them
2. mock datajoint so we don't attempt to connect to the DB
3. yield (test runs with mock DJ)
4. during teardown, restore the saved modules (from step 1) to sys.modules["aeon.dj_pipeline"] and any sys.modules["aeon"].dj_pipeline) - this keeps integration tests from seeing the stale mock

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -27,31 +27,66 @@ if "DJ_PORT" not in os.environ:
     os.environ["DJ_PORT"] = "3306"
 
 
-def pytest_configure(config):
-    """Called before test collection begins.
+def _make_mock_dj():
+    from unittest.mock import MagicMock
 
-    For unit tests, mock datajoint to prevent DB connection attempts
-    during import of aeon.dj_pipeline modules.
+    mock_dj = MagicMock()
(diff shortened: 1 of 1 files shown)


esutlie (approved):
Looks good to me!

ttngu207 (approved):

PR #573, opened 2026-05-15
Redirect DataJoint attach downloads to tmp dir in tests and restrict populate calls to experiment being tested

lochhh:
When running integration tests for OnixImuChunk, two related issues cause .joblib files to accumulate in cwd:
- OnixImuChunk.make() calls fetch1("sync_model"), which extracts .joblib files to dj.config["download_path"] (default is cwd) and these are never cleaned up after running the test suite
- All populate() calls are unrestricted. Since MySQL container is session-scoped, EphysSyncModel rows inserted by one test are visible to populate() calls in other tests, causing make() to process rows from unrelated experiments (created in other tests). This causes unnecessary attachment extractions from other tests' data.

This PR adds 2 fixes:
- the `dj_download_to_tmp` autouse fixture that overrides dj.config["download_path"], redirecting all attachment extractions to a per-test temporary directory that is cleaned up automatically
- restrict populate() calls to the experiment within the test, to ensure test isolation

### DIFF
--- a/tests/dj_pipeline/conftest.py
+++ b/tests/dj_pipeline/conftest.py
@@ -8,11 +8,30 @@ Test commands:
 
 import logging
 import os
+import tempfile
 from pathlib import Path
 
 import pytest
 
 logger = logging.getLogger(__name__)
+
+
+@pytest.fixture(autouse=True)
+def dj_download_to_tmp(request):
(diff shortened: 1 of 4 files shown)


esutlie (approved):
Looks great!

ttngu207 (approved):

Issue #584, opened 2026-06-02
`EphysChunk.ingest_chunks` fails when a probe_type has multiple electrode configs


AdrianRoggenbach:
### Problem
`EphysChunk.ingest_chunks()` cannot handle more than one `ElectrodeConfig` for a `probe_type`. At [ephys.py#L439](https://github.com/SainsburyWellcomeCentre/aeon_mecha/blob/main/aeon/dj_pipeline/ephys.py#L439) it fetches every config for the chunk's `probe_type` and raises `ValueError: Multiple electrode configs for {probe_type}` if more than one exists, because nothing records which configuration each recording actually used.

A single Neuropixels 2.0 probe (`probe_type = "neuropixels2.0"`) is routinely used with several recording configurations (different shanks/banks/depths), each registered as a separate `ElectrodeConfig` with a user-chosen name (e.g. `4shank-1000-1700um`, `3shank_2500-3200um`). All are valid and coexist, so the ingest step cannot proceed.

### Available information
Each epoch's `Metadata.yml` records the ProbeInterface file used per probe: `Devices → NeuropixelsV2e → ConfigurationA|B → ProbeInterfaceFileName` (e.g. `M82_ProbeA_4Shanks_1000_to_1700_um.json`). Currently this file name can only be informally stored in the free-text `electrode_config_description`, so chunks cannot be matched to their configuration file in a structured way.

### Proposed fix
1. Add a structured `config_file_name` attribute to `ElectrodeConfig` (the ProbeInterface JSON) with `unique index (probe_type, config_name)`.
2. Add entries manually when registering configs.
3. In `ingest_chunks`, resolve each chunk's config from its epoch+probe via `Metadata.yml` and look it up by `config_file_name`.
4. Fix the downstream channel-map step: `EphysBlockInfo.make` builds its channel↔electrode mapping via `load_device_channel_map(epoch_path)`, which globs the *epoch directory* for a ProbeInterface JSON — but that JSON isn't there (it lives in `recording_configurations/`), so it raises `FileNotFoundError` and can't pick the right config anyway. Make it load the config file referenced by the chunk's `ElectrodeConfig` (the new `config_file_name`, resolved under `recording_configurations/`) instead.

*Co-created with Claude Opus 4.8 🤖*

lochhh:
Closed in #585

Issue #583, opened 2026-06-02
Model `ephys.EphysEpoch` as a peer of `acquisition.Epoch` (symmetric epoch design for behaviour vs ephys)


AdrianRoggenbach:
## Summary
Behaviour and electrophysiology are recorded by two independent acquisition systems, but
the pipeline treats the ephys epoch as a *child* of the behaviour epoch
(`EphysEpoch -> acquisition.Epoch`). This issue proposes making `EphysEpoch` a **peer** of `acquisition.Epoch` — both direct
children of `acquisition.Experiment` — and mirroring the behaviour side's sub-tables on
the ephys side (`EphysEpochConfig`, `EphysEpochEnd`, `EphysChunk`).

## Background: two independent acquisition systems
| Modality | Machine | Directory type | Clock | Epoch = on/off of … |
|---|---|---|---|---|
| Behaviour | AEON3 | `raw` | HARP (master) | the HARP/Bonsai rig |
| Ephys | AEONX1 | `raw-ephys` | ONIX → HARP via `EphysSyncModel` | the ONIX/Neuropixels rig |

The two rigs start and stop independently; their epoch boundaries (directory timestamps)
do **not** coincide. They are aligned only after the fact through the shared HARP clock.

## Current design (asymmetric)
`acquisition.Epoch.ingest_epochs()` ingests *both* directory types into one table
(`acquisition.py`, the `raw` branch and the `raw-ephys` branch ~L229–345); `EphysEpoch`
then hangs one level below it:

```
                       acquisition.Experiment
                                │
                                ▼
                       acquisition.Epoch                    ← ONE table holds BOTH
                (raw + raw-ephys, keyed by directory_type)     behaviour & ephys epochs
          ┌────────────┬───────────┬──────────────────────────────────┐
          ▼            ▼           ▼                                   ▼
    EpochConfig    EpochEnd      Chunk                        ephys.EphysEpoch   ← ephys = GRANDCHILD
     (Imported)    (Manual)     (Manual)                          (Imported)
                                                                       │
                    ┌────────────────────┬────────────────┬───────────┴────────┐
                    ▼                    ▼                ▼                     ▼
               EphysSyncModel        EphysChunk       EphysBlock           OnixImuChunk
                 (Manual)             (Manual)         (Manual)            (Imported)
                                          │                │
                                          ▼                ▼
                                  (File / SyncModel    EphysBlockInfo
                                      parts)             (Imported)
```

[...]

Issue #581, opened 2026-06-01
`ephys.OnixImuChunk` populate call does not find IMU data


willreith:
When calling `.populate()` on `ephys.OnixImuChunk` on the `abcGolden01` dataset, all chunks have a sample count of 0 with corresponding empty `stream_df` dataframes. 

The issue seems to arise from `locate_bno055_chunk_index` in the `utils.onix_imu` module. When this is called during the populate call, it fails to find the BNO055 clock file starting with the ONIX timestamp that matches the expected `onix_ts_start` from the `ephys.EphysSyncModel` table. In fact, it seems like the expected timestamps are not found anywhere in the clock files.

ttngu207:
Tracked this down — root cause is a buried assumption in the old `locate_bno055_chunk_index` helper.

**What it was doing wrong**

The helper assumed that for every HarpSync sync window (one per hourly `*_HarpSync_*.csv`), there's exactly one `Bno055_Clock_N.bin` file whose **first sample** equals that window's `onix_ts_start`. So it just globbed the Bno055 files and exact-matched.

But HarpSync CSVs and Bno055 binaries come from two independent producers writing on different cadences:

- **HarpSync CSVs** are emitted by the Bonsai workflow on the host, on a wall-clock hourly cadence (you can see this in the filename: `_HarpSync_2026-05-11T07-00-00.csv`). Within each CSV, `clock_start` is the *first HARP-tick* that fell into that hour.
- **`Bno055_Clock_N.bin`** files are flushed by the ONIX firmware on the device, on a firmware-determined cadence (~10 min per file in the abcGolden01 recording).

There's no architectural reason for these boundaries to coincide — they're two independent timelines partitioning the same ONIX clock on different boundaries. On the golden dataset, **zero** of the 3 HarpSync `clock_start` values match any of the 12 Bno055 chunks' first samples. The lookup was searching for an alignment that physically can't exist on real data.

**Why this only surfaced now**

Earlier ephys recordings appear to have fit within a single wall-clock hour, so the producer always emitted one HarpSync CSV. With one sync window and one (or any) Bno055 chunk, the failing lookup either coincidentally succeeded or silently fell back to an empty-row insert that no assertion challenged. The `abcGolden01` recording is the first golden dataset that crosses hour boundaries (~35 min spanning 7:50 → 8:25, producing 3 HarpSync CSVs and 12 Bno055 chunks), which is what exposed the mismatch.

[...]

PR #585, opened 2026-06-05
Ephys pipeline restructure + IMU clock-alignment fix (#583 + #584 + #581) + PR 579 review fixes

ttngu207:
Combined work across four issues. Focused recap below; deeper writeups linked per issue.

## What & why

**PR 579 review fixes** — bugs and lint surfaced when running the ephys integration suite against the real golden dataset. Three baseline failures (F1, F2, F3) turned out to be one fixture bug (under-populated `ElectrodeConfig`) and two assertion bugs (wrong semantic on probe count and channel indices). Plus a safety guard on `_drop_test_schemas`, a monkey-patch leak fix, lint cleanup, and a `ctx` composite fixture.

**#583 — `EphysEpoch` as peer of `acquisition.Epoch`.** Behavior and ephys are independent acquisition systems with independent clocks. The schema previously modeled `EphysEpoch` as a child of `acquisition.Epoch`, conflating the two. They're now peers under `acquisition.Experiment`, each with parallel sub-tables (`EphysEpochConfig`, `EphysEpochEnd`). `EphysEpoch.epoch_start` is HARP-native (observed from the first HarpSync CSV), so cross-modal joins agree on one clock.

**#584 — `EphysChunk.ingest_chunks` handles multiple ElectrodeConfigs per probe_type.** A single Neuropixels 2.0 probe is routinely re-banked (different shanks/depths) across recordings, each producing a separate `ElectrodeConfig`. The old lookup errored on >1 config. Now `EphysEpochConfig.Insertion` carries a per-(epoch, probe) FK to the active `ElectrodeConfig` plus the source JSON basename, set at `EphysEpochConfig.populate` time. Chunk-ingest no longer reads `Metadata.yml` (resolution happens once, at populate). `EphysBlockInfo` also gained a uniform-config check — refuses to populate a block spanning two configs instead of silently picking one.

**#581 — IMU clock-alignment.** `OnixImuChunk` produced empty rows on real data because the old lookup assumed each HarpSync sync window's `onix_ts_start` equals the first sample of one Bno055 binary chunk. The two are independent producers on different cadences — that alignment can't exist on real data. New overlap-based selector finds all Bno055 chunks intersecting a window, concatenates, and filters. The previous synthetic test masked the bug by constructing perfectly aligned fixtures; that's been replaced with realistic staggered data. Full root-cause writeup posted to #581.

## Test coverage

**Unit: 102 passing** (was 81 baseline, +21):
- New helpers (`create_electrode_config`, `parse_metadata_probe_configs`, `resolve_epoch_probe_json`, `find_overlapping_bno055_chunks`, refactored `load_device_channel_map`) all have edge-case coverage.
- 2 new end-to-end pattern tests for the IMU overlap → concat → filter pipeline on staggered synthetic data.

**Full integration: 65 passing, 15 skipped, 0 failed.**
- Behavior golden dataset (`test_full_ingestion.py`): all pass.
- Ephys golden dataset (`test_ephys_ingestion.py`): all pass, including 5 new tests added for this PR — `EphysSyncModel` row-count/r²/monotonicity (3) and `OnixImuChunk` on golden data (2).
- New synthetic integration tests for `EphysEpochEnd` look-back behavior and `EphysBlockInfo` multi-config validation.
- ONIX IMU, codec, acquisition environment, load_metadata: all pass.
- The 15 skips are the spike-sorting cascade tests gated on the `golden_test_sorting/` directory (not yet uploaded to S3) + a few module-level skips for missing optional deps.

**Lint:** ruff clean on all touched files.

## Migration

Breaking change to the ephys schema (drop and recreate; re-ingest from raw). No row-level migration script.

## Out of scope (deferred)

- Production-data migration scripts.
- Cross-modal alignment helper (behavior epoch ↔ ephys epoch).
- Setup-script (`ephys_v2_setup.py`, `ephys_mock_ingestion.py`) modernization.

## Docs

- New `docs/specs/SPEC_EPHYS_PIPELINE.md` — architecture, ingestion order, HARP-clock semantics.
- Updated `docs/specs/SPEC_EPHYS_TESTING.md`.


AdrianRoggenbach:
Looks very good, thanks! A few minor comments:

- In EphysEpochConfig, the field has_ephys ([here](https://github.com/SainsburyWellcomeCentre/aeon_mecha/blob/tn/ephys-pipeline-restructure/aeon/dj_pipeline/ephys.py#L209)) is not necessary anymore, since no behavior-only epochs are passed through the ephys tables anymore.
- In ElectrodeConfig ([LINK](https://github.com/SainsburyWellcomeCentre/aeon_mecha/blob/tn/ephys-pipeline-restructure/aeon/dj_pipeline/ephys.py#L60)), it might be nicer to have the name of the file as an attribute, that one can choose a shorter primary key in case of a long config file name. This would make it also more explicit that this is a reference to a file name, which can then also include the .json extension.
- Since we need to drop all ephys, spike_sorting and spike_sorting_curation tables to implement this change, it would be a good opportunity to rename all variables across these three files according to DJ2 preference ([LINK](https://docs.datajoint.com/reference/specs/type-system/#native-passthrough-types)). E.g. int -> int32, float -> float32 or 64, ...

esutlie (commented):
@ttngu207 This looks great to me! I agree with Adrian's suggestions. We could also consider dropping `ephys_v2_setup.py` and `ephys_v2_validate.py`, since they predate, and are essentially superseded by, the tests.

ttngu207:
Thanks Adrian — addressed in the latest commits:

- `has_ephys` dropped; the 4 silent-fail paths in `EphysEpochConfig.make()` now raise diagnostic exceptions instead. `n_probes` kept as a useful summary counter.
- DJ 2.x native types renamed across `ephys.py`, `spike_sorting.py`, `spike_sorting_curation.py` (`int` → `int32`, `float` → `float32`, `bigint` → `int64`, `smallint` → `int16`).
- On the ElectrodeConfig file name: `EphysEpochConfig.Insertion.config_file_name` already records that per (epoch, probe), so I left ElectrodeConfig alone to avoid duplication. Let me know if you'd prefer it stored in both places.

Also dropped `ephys_v2_setup.py`, `ephys_v2_validate.py`, and `ephys_mock_ingestion.py` per Elissa's note.

esutlie (approved):

esutlie:
These changes all look good!

PR #587, opened 2026-06-11
Add an xarray NetCDF codec for storing behavioral data in DataJoint

AdrianRoggenbach:
### Overview

We want to process behavioral data with [`movement`](https://movement.neuroinformatics.dev/), which represents pose/behavioral tracks natively as `xarray.Dataset`s. xarray gives us labeled, multi-dimensional arrays (time × channel, time × keypoint × space, …) with coordinates and attributes carried alongside the data.

A key benefit is **lazy loading along the time dimension**. Because each recording is stored as a NetCDF-4 file and reopened lazily, behavioral data can be concatenated along the `time` axis without pulling the whole array into RAM — only the time coordinate is materialized, and the heavy data variables stay on disk until a slice is actually accessed. This lets us stitch many recordings/segments into one logical timeline and work with it as if it were in memory, while keeping the actual memory footprint to the time axis plus whatever window you touch.

This PR adds the storage layer that makes that possible: a DataJoint codec that round-trips an `xarray.Dataset` to and from a NetCDF file.

### What's added

- **`XArrayNetCDFCodec`** (`aeon/dj_pipeline/utils/xarray_codec.py`) — a `SchemaCodec` modeled on DataJoint's built-in `NpyCodec`. It persists an `xarray.Dataset` as a single NetCDF-4 `.nc` file at a schema-addressed, tokened path (`{schema}/{table}/{pk}/{field}_<token>.nc`) and reopens it lazily on fetch. Registered in `aeon/dj_pipeline/__init__.py` so it's available as the `<xarray@store>` column type.
- **`h5netcdf[h5py]`** added to `pyproject.toml` dependencies (with `uv.lock` updated) — the NetCDF engine used for read/write.
- **Tests** — unit tests for the pure xarray ⇄ NetCDF on-disk contract (no DB), and integration tests against a real DataJoint table on a testcontainers MySQL instance with a temp-dir file store.

### Behavior / design notes

- **Lazy by default, no dask.** Fetch returns a lazy `Dataset` opened with `chunks=None` (xarray's own lazy indexing). Data variables stay on disk until accessed; `.load()` realizes them into numpy. This is what enables the time-axis-only memory profile described above.
- **`Dataset` only.** A `DataArray` is rejected with a hint to call `.to_dataset()` first, keeping insert and fetch symmetric (you never store a `DataArray` and get a `Dataset` back).
- **Generic.** The codec knows nothing about any particular dataset schema or domain format — it just round-trips an `xarray.Dataset`.
- **File stores only.** For `protocol: file` stores it writes/opens directly by local path rather than buffering bytes; it asserts `protocol == "file"` and does not support object stores (S3).

### Known limitations

- File-store only — no S3/object-store support yet.
- The xarray netCDF format only supports timezone-naive datetimes.

Fixed in newest DataJoint version: `dj.gc.collect()` is unsafe for schema-addressed codecs in DataJoint 2.2.2 (it can corrupt the store); use orphan-file cleanup instead of the built-in GC for these columns.


AdrianRoggenbach:
Since the issue https://github.com/datajoint/datajoint-python/issues/1469 has been closed on datajoint-python, I have now integrated the tests for the garbage collector, and they all pass.

If datajoint is installed from GitHub, the codec should now work as expected.

aeon/dj_pipeline/utils/xarray_codec.py
```diff
@@ -0,0 +1,81 @@
...
+    def encode(self, value: xr.Dataset, *, key: dict | None = None, store_name: str | None = None) -> dict:
+        """Write the Dataset to a NetCDF-4 file and return JSON metadata."""
+        schema, table, field, primary_key = self._extract_context(key)
+        config = (key or {}).get("_config")
+        path, _token = self._build_path(
+            schema, table, field, primary_key, ext=".nc", store_name=store_name, config=config
+        )
+        local_path = self._local_path(path, store_name, config)
+        os.makedirs(os.path.dirname(local_path), exist_ok=True)
+        value.to_netcdf(local_path, engine="h5netcdf")
```
lochhh: is there a particular reason for preferring h5netcdf to netcdf4, which xarray recommends? the other alternative would be to install [`xarray[io]`](https://docs.xarray.dev/en/latest/getting-started-guide/installing.html#for-netcdf-and-io) and leave `engine` unspecified, letting xarray decide based on the  current order (netCDF4, h5netcdf, scipy)

aeon/dj_pipeline/utils/xarray_codec.py
```diff
@@ -0,0 +1,81 @@
...
+        return {
+            "path": path,
+            "store": store_name,
+            "dims": dict(value.sizes),
+            "data_vars": list(value.data_vars),
+            "provenance": {"datajoint-python": dj.__version__},
```
lochhh: not sure this is useful
```suggestion
```

aeon/dj_pipeline/utils/xarray_codec.py
```diff
@@ -0,0 +1,81 @@
...
+        }
+
+    def decode(self, stored: dict, *, key: dict | None = None) -> xr.Dataset:
+        """Reopen the stored NetCDF file as a lazy xarray.Dataset (no dask)."""
+        config = (key or {}).get("_config")
+        local_path = self._local_path(stored["path"], stored.get("store"), config)
+        return xr.open_dataset(local_path, engine="h5netcdf", chunks=None)
```
lochhh:
```suggestion
        return xr.open_dataset(local_path, engine="h5netcdf")
```

aeon/dj_pipeline/__init__.py
```diff
@@ -44,6 +44,9 @@ def _escape_dict_as_json(val, charset, mapping=None):
     AeonStreamCodec,
     OnixStreamCodec,
 )
+from aeon.dj_pipeline.utils.xarray_codec import (  # pyright: ignore[reportUnusedImport]
+    XArrayNetCDFCodec,
```
lochhh, line 46: this could be merged into codec.py, just need to rephrase the docstrings there

aeon/dj_pipeline/utils/xarray_codec.py
lochhh: both h5netcddf, netcdf4 engines support compression by specifying e.g. `encoding={"zlib": True, ...}`, is this something we would want to consider?

lochhh:
**[IGNORE]: I was on DataJoint 2.3.1** 😅 

`uv run pytest .\tests\dj_pipeline\utils\test_xarray_codec_integration.py` failed on Windows with:

> FAILED tests/dj_pipeline/utils/test_xarray_codec_integration.py::TestGarbageCollection::test_live_rows_leave_nothing_orphaned - assert 2 == 0
> FAILED tests/dj_pipeline/utils/test_xarray_codec_integration.py::TestGarbageCollection::test_dry_run_reports_the_orphan_but_deletes_nothing - assert (1, 2) == (1, 1)
> FAILED tests/dj_pipeline/utils/test_xarray_codec_integration.py::TestGarbageCollection::test_collect_reclaims_only_the_orphaned_file - assert 2 == 1
> FAILED tests/dj_pipeline/utils/test_xarray_codec_integration.py::TestGarbageCollection::test_collect_is_idempotent - FileNotFoundError: [Errno 2] Unable to synchronously open file (unable to open file: name = '...\pytest-766\test_collect_is_ide...

Raised https://github.com/datajoint/datajoint-python/issues/1520

aeon/dj_pipeline/utils/xarray_codec.py
AdrianRoggenbach: Switching now to the recommended engine netcdf4. For now, I would leave this codec without compression and later add a codec with compression if we end up with very large files.

aeon/dj_pipeline/utils/xarray_codec.py
AdrianRoggenbach: chunks=None is the default behavior of this function. Removing it as suggested.

aeon/dj_pipeline/__init__.py
AdrianRoggenbach: This is done in #8a4d2da

AdrianRoggenbach:
I have addressed the remaining comments from @lochhh by changing the codec to NetCDF4 and moving the codec from its own file into `utils/codec.py`.

AdrianRoggenbach:
Removed the temporary fix to install datajoint from GitHub master branch, and replaced it with the newest DataJoint version 2.3.2 which includes the fixes for the garbage collector.

This version pin also makes the pull request #600 not necessary anymore.

ttngu207 (approved):

PR #589, opened 2026-06-16
Make zarr the default format for spike sorting intermediates

esutlie:
Switches the spike sorting pipeline's default intermediate format from uncompressed binary to zarr with Blosc-zstd compression. PreProcessing now writes `recording.zarr` instead of `recording.dat`, and PostProcessing writes `sorting_analyzer.zarr`. The compression is lossless and built into SpikeInterface, so there are no new dependencies.

On the golden data this is roughly a 60% reduction in on-disk size (recording 12.7 GB to 5.0 GB, sorting analyzer 0.5 GB to 0.3 GB), and pipeline I/O is faster because less data moves over the Ceph network.

Binary is still fully supported. The format is controlled by a `save_format` key in each paramset's params, and this PR only changes the default from "binary" to "zarr". Paramsets that explicitly set `save_format: "binary"` behave exactly as before.

Main changes:
- Default flipped to "zarr" in PreProcessing, SpikeSorting, PostProcessing, and ApplyOfficialCuration.
- Non-numeric recording properties (structured arrays like `contact_vector`, string and object arrays) are stripped before the zarr write, because zarr v2 cannot serialize them without an explicit object codec.
- SpikeInterface appends `.zarr` to the analyzer folder name, so a shared `_resolve_analyzer_dir` helper resolves the analyzer path for all consumers, and PostProcessing returns the actual `.zarr` path so its files register correctly.
- Zarr chunk files are excluded from the File part table, since a zarr directory is thousands of small files.
- The two pure helpers live in `aeon/dj_pipeline/utils/spike_sorting_utils.py` so they can be unit tested without a database connection.
- Adds `docs/specs/SPEC_ZARR_INTERMEDIATES.md` describing the approach and measured results.

Backward compatible: `si.load()` reads both formats, existing binary intermediates are untouched, and there is no database migration. Old data does not need to be converted.

Validated on the SWC HPC against the golden dataset: the full unit suite passes and the ephys golden integration suite is 29/29, including new assertions for the zarr recording and analyzer.

### DIFF
--- a/aeon/dj_pipeline/spike_sorting.py
+++ b/aeon/dj_pipeline/spike_sorting.py
@@ -565,28 +565,22 @@ class PostProcessing(dj.Computed):
 
         job_kwargs = postprocessing_params.get("job_kwargs", {"n_jobs": -1, "chunk_duration": "1s"})
 
-        analyzer_output_dir = output_dir / "sorting_analyzer"
-        for check_dir in [analyzer_output_dir, output_dir / "sorting_analyzer.zarr"]:
-            if check_dir.exists() and any(check_dir.iterdir()):
-                raise FileExistsError(
-                    f"Sorting analyzer directory already contains data: {check_dir}. "
-                    "Refusing to overwrite existing sorting data."
-                )
-
-        has_units = si_sorting.unit_ids.size > 0
-
-        # ---- run the analyzer extensions ----
-        if not has_units:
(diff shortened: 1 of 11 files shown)


lochhh (approved):
Thanks @esutlie !

ttngu207 (approved):

PR #590, opened 2026-06-19
Suggestions for PR #589

lochhh:
Review suggestions for #589. 

- Lint & formatting cleanup across changed files
  - consistent indentation, trailing commas, whitespace, docstrings moved to PEP 257 style
  - Modern type annotations: replace `Dict`, `List`, `Tuple`, `Optional` with `dict`, `list`, `tuple` builtins; add `Iterator` from `collections.abc`
  - Stricter iteration: add `strict=True/False` to `zip()` calls; replace bare `assert` with `ValueError` for runtime checks
  - Remove unused  `from spikeinterface import sorters, preprocessing` imports
- Public API for spike sorting utils: rename `_resolve_analyzer_dir` → `resolve_analyzer_dir` and `_strip_non_numeric_properties` → `strip_non_numeric_properties` (these are used across modules, not private)
- Deduplicate logic
  - move `skip_kilosort_preprocessing` above the `save_format` branch; 
  - simplify `PostProcessing` analyzer dir check to only verify the relevant format
  - use `Path.is_relative_to` for zarr filtering
- Raise early in`resolve_analyzer_dir`: add `FileNotFoundError` when neither binary nor zarr analyzer dir exists
- CI test coverage
  - add `spikeinterface` to the test dependency group so unit tests run in CI instead of being skipped
  - remove unnecessary `pytest.importorskip` guards (including `aeon.schema.ephys` guard — it's a local module, not external?)
- Document `save_format` parameter in `step03_spike_sorting.py` runbook (also noticed `"export_to_phy": False, "export_report": True,` are unused/ignored - should we remove these?)
- Update SPEC_ZARR_INTERMEDIATES.md to match code changes

@esutlie could you please run the full test suite with golden datasets to make sure nothing is broken?

### DIFF
--- a/aeon/dj_pipeline/spike_sorting.py
+++ b/aeon/dj_pipeline/spike_sorting.py
@@ -565,28 +565,22 @@ class PostProcessing(dj.Computed):
 
         job_kwargs = postprocessing_params.get("job_kwargs", {"n_jobs": -1, "chunk_duration": "1s"})
 
-        analyzer_output_dir = output_dir / "sorting_analyzer"
-        for check_dir in [analyzer_output_dir, output_dir / "sorting_analyzer.zarr"]:
-            if check_dir.exists() and any(check_dir.iterdir()):
-                raise FileExistsError(
-                    f"Sorting analyzer directory already contains data: {check_dir}. "
-                    "Refusing to overwrite existing sorting data."
-                )
-
-        has_units = si_sorting.unit_ids.size > 0
-
-        # ---- run the analyzer extensions ----
-        if not has_units:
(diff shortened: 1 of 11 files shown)


esutlie (approved):
Thanks for these improvements @lochhh! I ran the full test suite on your branch against the golden datasets on the HPC and it all passes. I agree with all your suggestions, let's get them merged in. I also wrote up how to run the golden tests yourself, here: https://github.com/SainsburyWellcomeCentre/aeon_mecha/blob/es/small-fixes/docs/running_golden_tests_on_hpc.md which will be included in an upcoming PR with a few other small fixes and improvements.

PR #593, opened 2026-06-26
test(ephys): close PR 592's coverage gap + repoint foragingabc dep to main

ttngu207:
## Summary
PR #592 fixed a real 1-second offset in the HARP↔ONIX sync regression (used `Value.HarpTime`, the integer second about-to-end; switched to the `Seconds` index column). All existing `EphysSyncModel` tests passed both before and after the fix — they only check internal-consistency (count, r², monotonicity), which a systematic offset doesn't disturb. This PR adds the missing coverage and tidies up the dep pin while we're here.

## Changes
- `tests/dj_pipeline/test_ephys_ingestion.py`: two new tests in `TestEphysSyncModel`, each using an **independent** HARP-time anchor:
  - **`test_sync_start_aligns_with_hour_boundary`** — for HarpSync CSVs whose hour bucket starts after `epoch_start` (08-00, 09-00 in golden), the first `Seconds` value lands within the first second of the hour, so `sync_start.minute == 0 and sync_start.second == 0`. Pre-PR 592 these would be 59/59.
  - **`test_sync_start_matches_seconds_column_not_value_harptime`** — re-reads one HarpSync CSV directly, derives what `sync_start` would be from `Seconds` vs `Value.HarpTime`, asserts the stored value matches the `Seconds`-derived one. Most direct test of the fix; works for any sync row.
- `pyproject.toml` / `uv.lock`: repoint `swc-aeon-rigs-foragingabc` from the temporary `tn/env-data-readers` branch (the upstream PR 585 staged on) to `main`. SainsburyWellcomeCentre/aeon_exp_foragingABC#83 merged on 2026-06-11. The lock refresh also bumped lxml 4.9→6.1, matplotlib 3.7→3.11, numpy 1.26→2.3, opencv-python 4.7→4.13 as a consequence of the new resolution graph.

## Why not a "first chunk_start ≥ dir timestamp" check
Initially planned this as a third test, but the golden data showed `chunk_start` legitimately precedes the dir wall-clock timestamp by ~23s (HARP logging starts before Bonsai provisions the dir). The dir name isn't a useful anchor for HARP-clock values.

## Test plan
- Integration suite: 123 passed, 10 skipped, 0 failed (was 121/10/0 on main; +2 from the new tests).
- Both new tests verified to fail if the PR 592 fix were reverted (the assertion messages explicitly reference what regression would look like).


esutlie (approved):
LGTM!

PR #597, opened 2026-07-09
Read compressed raw ephys files (prefer .zarr, fall back to .bin)

esutlie:
## Summary
- Add `resolve_ephys_file` in `utils/ephys_utils.py`. Given a raw amplifier `.bin` path, it returns the file to actually read on disk, preferring the compressed `.zarr` twin under the processed store, falling back to the raw `.bin`, and raising if neither exists.
- Wire it into `PreProcessing.make_compute` so chunk reads resolve on disk at load time. On the `.zarr` branch it re-applies channel gains and offsets after `si.load`, since the compression library writes zarr from a plain `read_binary` that carries no gains.
- Unit tests for the resolver, covering prefer-zarr, fall-back-to-bin, both/only-one/neither present, and the ambiguous multi-`raw` path case.
- Golden integration test confirming a `.zarr` twin read matches the `.bin` read byte-for-byte on real data.

## Context
Companion to the standalone `aeon_raw_compression` library, which losslessly compresses raw `*_AmplifierData_*.bin` files to `.zarr` under a processed data root that mirrors the raw sub-path. This PR is the read side, where aeon_mecha finds and reads those compressed files.

The design, locked with Thinh, is to store paths as read and never rewrite them. The resolver maps a stored raw path to the file on disk at load time by a fixed convention, swapping the `raw` path component for `processed` and `.bin` for `.zarr`. Compression is optional and per-user, so binary reading stays first-class and there is no dependency on `aeon_raw_compression`.

No database changes. No schema change, no migration, no stored-path rewrites. Old data with no `.zarr` twin reads exactly as before.

Verified with the unit suite locally (113 passed) and the ephys golden suite on the SWC HPC (32 passed), including the new equivalence test. The spec is at `docs/specs/SPEC_READ_COMPRESSED.md`.

## Note
Also includes a small unrelated docs fix removing a stale `uv pip install cryptography` step from `docs/running_golden_tests_on_hpc.md`, since `cryptography` became a core dependency in #591.

### DIFF
--- a/aeon/dj_pipeline/utils/ephys_utils.py
+++ b/aeon/dj_pipeline/utils/ephys_utils.py
@@ -21,6 +21,59 @@ DEVICE_PROBE_TYPE_MAP = {
 }
 
 
+def resolve_ephys_file(raw_bin_path: Path) -> Path:
+    """Resolve a raw amplifier ``.bin`` path to the file that exists on disk.
+
+    Compressed raw ephys (from the companion ``aeon_raw_compression`` library) is
+    written as a ``.zarr`` archive under a *processed* data root that mirrors the
+    raw sub-path: a file at ``.../data/raw/<sub>/<name>.bin`` has its compressed
+    twin at ``.../data/processed/<sub>/<name>.zarr``. This resolver prefers that
+    ``.zarr`` twin and falls back to the raw ``.bin`` (compression is optional and
+    per-user; either or both may be present).
+
+    The raw->processed mapping is a fixed convention, not configurable: replace
+    the single ``"raw"`` path component with ``"processed"`` and swap the suffix
+    to ``.zarr``. This mirrors ``aeon_raw_compression``'s path mapping
+    (``pipeline._processed_zarr_path``) WITHOUT importing that library, since not
+    all users install it -- keep the two in sync by hand if either changes.
+
+    Args:
+        raw_bin_path: Absolute path to the raw ``.bin`` amplifier file as
+            reconstructed by the pipeline (``data_directory / file_path``).
+
+    Returns:
+        The ``.zarr`` twin if it exists, else the raw ``.bin`` if it exists.
+
+    Raises:
+        ValueError: If the path has more than one ``"raw"`` component (the
+            processed-store root would be ambiguous).
+        FileNotFoundError: If neither the ``.zarr`` twin nor the ``.bin`` exists.
+    """
+    raw_bin_path = Path(raw_bin_path)
+    parts = raw_bin_path.parts
+
+    raw_indices = [i for i, part in enumerate(parts) if part == "raw"]
+    if len(raw_indices) > 1:
+        raise ValueError(
+            f"Cannot resolve compressed twin for {raw_bin_path}: path has "
+            f"{len(raw_indices)} 'raw' components; the processed-store root is "
+            f"ambiguous."
+        )
+
+    zarr_candidate = None
+    if len(raw_indices) == 1:
+        i = raw_indices[0]
+        zarr_candidate = Path(*parts[:i], "processed", *parts[i + 1 :]).with_suffix(".zarr")
+
+    if zarr_candidate is not None and zarr_candidate.exists():
+        return zarr_candidate
+    if raw_bin_path.exists():
+        return raw_bin_path
+
+    checked = str(raw_bin_path) if zarr_candidate is None else f"{zarr_candidate} or {raw_bin_path}"
+    raise FileNotFoundError(f"No ephys file found (checked {checked}).")
+
+
 def get_probe_id(metadata: dict | None, device_name: str, probe_label: str) -> str | None:
     """Extract probe identifier from metadata.
 
--- a/aeon/dj_pipeline/utils/ephys_utils.py
+++ b/aeon/dj_pipeline/utils/ephys_utils.py
@@ -52,17 +52,15 @@ def resolve_ephys_file(raw_bin_path: Path) -> Path:
     raw_bin_path = Path(raw_bin_path)
     parts = raw_bin_path.parts
 
-    raw_indices = [i for i, part in enumerate(parts) if part == "raw"]
-    if len(raw_indices) > 1:
+    if parts.count("raw") > 1:
         raise ValueError(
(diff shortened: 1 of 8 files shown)


ttngu207:
@esutlie thank you. LGTM. This PR is good to merge. Would you bump version here before we merge?

esutlie:
Thanks @ttngu207, I've bumped the version now 👍

aeon/dj_pipeline/utils/ephys_utils.py
```diff
@@ -21,6 +21,59 @@
...
+    if len(raw_indices) > 1:
+        raise ValueError(
+            f"Cannot resolve compressed twin for {raw_bin_path}: path has "
+            f"{len(raw_indices)} 'raw' components; the processed-store root is "
+            f"ambiguous."
+        )
+
+    zarr_candidate = None
+    if len(raw_indices) == 1:
+        i = raw_indices[0]
```
lochhh:
```suggestion
    if parts.count("raw") > 1:
        raise ValueError(
            f"Cannot resolve compressed twin for {raw_bin_path}: path has "
            f"multiple 'raw' components; the processed-store root is ambiguous."
        )

    zarr_candidate = None
    if "raw" in parts:
        i = parts.index("raw")
```

tests/dj_pipeline/utils/test_ephys_utils_unit.py
```diff
@@ -632,3 +632,86 @@ def test_raises_on_no_active_contacts(self, tmp_path):
...
+        # "raw-ephys" is a DataJoint directory_type label, never a physical path
+        # component. Exact-component matching must not treat it as the "raw" store
+        # root, so no .zarr twin is inferred and we fall back to the .bin.
+        from aeon.dj_pipeline.utils.ephys_utils import resolve_ephys_file
+
+        bin_path = tmp_path / "raw-ephys" / "AEONX1" / "f.bin"
+        bin_path.parent.mkdir(parents=True)
+        bin_path.write_bytes(b"\x00")
+
+        assert resolve_ephys_file(bin_path) == bin_path
```
lochhh, line 710: Refactor `raw_bin_path` `zarr_path` as fixtures so we don't need to pass `tmp_path`
```suggestion
class TestResolveEphysFile:
    """Pure path resolver: prefer the .zarr twin under the processed store, else the raw .bin, else error.

    No DB. See SPEC_READ_COMPRESSED.md.
    """

    @pytest.fixture
    def raw_bin_path(self, tmp_path):
        """Return a raw .bin path under tmp_path with a 'raw' component."""
        # tmp_path has no "raw" component, so the one we add is unambiguous.
        bin_path = tmp_path / "raw" / "AEONX1" / "exp" / "dev" / "Probe_AmplifierData_0.bin"
        bin_path.parent.mkdir(parents=True, exist_ok=True)
        return bin_path

    @pytest.fixture
    def zarr_path(self, tmp_path):
        """Return a processed .zarr path under tmp_path."""
        return tmp_path / "processed" / "AEONX1" / "exp" / "dev" / "Probe_AmplifierData_0.zarr"

    def test_prefers_zarr_when_both_present(self, raw_bin_path, zarr_path):
        from aeon.dj_pipeline.utils.ephys_utils import resolve_ephys_file

        raw_bin_path.write_bytes(b"\x00\x00")
        zarr_path.mkdir(parents=True)  # zarr is a directory on disk
        # Pins the raw->processed mapping AND the zarr preference.
        assert resolve_ephys_file(raw_bin_path) == zarr_path

    def test_falls_back_to_bin_when_only_bin(self, raw_bin_path):
        from aeon.dj_pipeline.utils.ephys_utils import resolve_ephys_file

        raw_bin_path.write_bytes(b"\x00\x00")
        assert resolve_ephys_file(raw_bin_path) == raw_bin_path

    def test_returns_zarr_when_only_zarr(self, raw_bin_path, zarr_path):
        from aeon.dj_pipeline.utils.ephys_utils import resolve_ephys_file

        # raw_bin_path's parent dir exists, but no .bin file
        zarr_path.mkdir(parents=True)
        assert resolve_ephys_file(raw_bin_path) == zarr_path

    def test_raises_when_neither_exists(self, raw_bin_path):
        from aeon.dj_pipeline.utils.ephys_utils import resolve_ephys_file

        # nothing created beyond raw_bin_path's parent dir
        with pytest.raises(FileNotFoundError, match=r"processed.*\.zarr.*\.bin"):
            resolve_ephys_file(raw_bin_path)

    def test_raises_on_multiple_raw_components(self, tmp_path):
        from aeon.dj_pipeline.utils.ephys_utils import resolve_ephys_file

        bin_path = tmp_path / "raw" / "sub" / "raw" / "f.bin"
        bin_path.parent.mkdir(parents=True)
        bin_path.write_bytes(b"\x00")
        with pytest.raises(ValueError, match="multiple 'raw' components"):
            resolve_ephys_file(bin_path)

    def test_no_raw_component_falls_back_to_bin(self, tmp_path):
        from aeon.dj_pipeline.utils.ephys_utils import resolve_ephys_file

        bin_path = tmp_path / "data" / "f.bin"  # no "raw" component
        bin_path.parent.mkdir(parents=True)
        bin_path.write_bytes(b"\x00")
        assert resolve_ephys_file(bin_path) == bin_path

    def test_raw_ephys_component_not_treated_as_raw(self, tmp_path):
        # "raw-ephys" is a DataJoint directory_type label, never a physical path
        # component. Exact-component matching must not treat it as the "raw" store
        # root, so no .zarr twin is inferred and we fall back to the .bin.
        from aeon.dj_pipeline.utils.ephys_utils import resolve_ephys_file

        bin_path = tmp_path / "raw-ephys" / "AEONX1" / "f.bin"
        bin_path.parent.mkdir(parents=True)
        bin_path.write_bytes(b"\x00")
        assert resolve_ephys_file(bin_path) == bin_path
```

lochhh (dismissed):
Thanks @esutlie LGTM too! Left minor suggestions for simplification.

esutlie:
Thanks @lochhh! All good suggestions, I've incorporated them in the new commits. Let me know if there are any other improvements before we merge!

ttngu207 (approved):

PR #599, opened 2026-07-15
Fix #598: clean up sync_model .joblib temp files left behind by attach fetches

### LINKED ISSUE #598: ingest_chunks creates temporary .joblib files but does not delete them
After running `ephys.EphysChunk.ingest_chunks(experiment_name)`, there are many `NeuropixelsV2_HarpSync_2026-06-25T130000Z.joblib` files in the base directory that the script was run from.

It would be good to save them in a specified folder for temporary files and delete them when they are not used anymore.

esutlie:
Fixes #598.

After running `EphysChunk.ingest_chunks`, `.joblib` files (`NeuropixelsV2_HarpSync_*.joblib`) were piling up in whatever directory the script was run from and never getting cleaned up.

The cause is that `EphysSyncModel.sync_model` is an `<attach>` column. DataJoint materializes `<attach>` values to a file under `dj.config["download_path"]` (the current working directory by default) on every fetch, and it deliberately never deletes them. So every fetch of a sync_model row drops a `.joblib` into the working directory. `ingest_chunks`, `OnixImuChunk.make`, and `OnixImuChunk.synced_df` all fetch sync_model and were leaking these files. `SyncedSpikes.make` in spike_sorting already avoided this by wrapping its fetch in a temp download dir.

This adds a small `attach_download_tempdir()` context manager in `ephys_utils` that points `download_path` at a `TemporaryDirectory` for the duration of the fetch-and-load, so the extracted files land in a temp dir that gets removed on exit. Every sync_model fetch site now uses it, and `SyncedSpikes.make` is refactored to the shared helper instead of its own inline copy.

This is a targeted fix that keeps the `<attach>` column as-is. The deeper options (storing the regression as a blob, json, or a custom codec, or just storing the slope and intercept) all eliminate the files at the root, but they need a one-time backfill of already-shipped data, so I'm holding those for a separate discussion with Thinh.

I also added unit coverage for the helper and an integration test that drives `ingest_chunks`, `OnixImuChunk.make`, and `synced_df` against a real attach round-trip and asserts nothing is left behind. A negative-control run with the helper disabled fails those tests and reproduces the exact `HarpSync_*.joblib` leak, so they guard the real behavior. Verified on the HPC against aeondj.

### DIFF
--- a/aeon/dj_pipeline/ephys.py
+++ b/aeon/dj_pipeline/ephys.py
@@ -1,6 +1,7 @@
 """DataJoint schema for the ephys pipeline."""
 
 import re
+import tempfile
 from datetime import datetime
 from pathlib import Path
 from typing import Any
@@ -406,8 +407,6 @@ class EphysSyncModel(dj.Manual):
         Args:
             experiment_name: Name of the experiment to process
         """
-        import tempfile
-
         import joblib
 
         from aeon.schema.ephys import social_ephys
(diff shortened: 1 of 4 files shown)


aeon/dj_pipeline/utils/ephys_utils.py
```diff
@@ -5,15 +5,39 @@
...
 from datetime import datetime
 from pathlib import Path, PureWindowsPath
 
 import datajoint as dj
 
 logger = dj.logger
 
+
+@contextlib.contextmanager
+def attach_download_tempdir():
```
ttngu207: Do we need this to be a helper function? How about just use `with tempfile.TemporaryDirectory() as tmpdir, dj.config.override(download_path=tmpdir):` inline, directly?

esutlie: That's cleaner I think. I've switched it in all four places, thanks @ttngu207!

aeon/dj_pipeline/ephys.py
```diff
@@ -867,12 +873,16 @@ def synced_df(cls, key):
         For raw ONIX-clock-indexed data, fetch ``stream_df`` directly instead
         of using this helper.
         """
+        import tempfile
```
ttngu207: I think `tempfile` is safe as a top level import, no need for lazy import here. Unless you're concerned with circular import or version/dependency conflicts?

ttngu207:
Thanks, looks good @esutlie , version bump here too? And one tiny nitpick on `tempfile` as top level import (see above)

esutlie:
Awesome! I addressed the tempfile import fix and added a version bump. After #597 is merged I will rebase here and run the tests again before marking this ready to merge. Thanks @ttngu207!

ttngu207 (dismissed):

esutlie:
@ttngu207 I rebased on main and bumped the version. Please approve and merge if it all looks good, whenever you get the chance!

ttngu207 (approved):

PR #600, opened 2026-07-17
Restore tracebacks for uncaught exceptions

AdrianRoggenbach:
Importing DataJoint installs a process-wide sys.excepthook that routes every uncaught exception through its logger with exc_info attached. Its LevelAwareFormatter overrides Formatter.format and returns only record.getMessage(), never calling formatException, so the traceback is attached to the record and then discarded.

Every crash in any script importing aeon.dj_pipeline therefore prints a single line, with no exception type, message, or traceback:

    [2026-07-17 13:51:31][ERROR]: Uncaught exception

Neither DJ_LOG_LEVEL nor DJ_LOG_STREAM affects this, so it cannot be worked around by configuration.

Reinstate the excepthook with the traceback rendered into the message itself, which every formatter emits unconditionally. DataJoint's excepthook is its only exc_info call site, so formatting the text here keeps the patch independent of the formatter chain and free of double-rendering if DataJoint later fixes LevelAwareFormatter. KeyboardInterrupt still defers to `sys.__excepthook__`.

### DIFF
--- a/aeon/dj_pipeline/__init__.py
+++ b/aeon/dj_pipeline/__init__.py
@@ -3,6 +3,8 @@
 import json
 import logging
 import os
+import sys
+import traceback
 from typing import cast
 
 import pymysql.converters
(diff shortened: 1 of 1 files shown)


dimitri-yatsenko:
This was fixed in v2.3.2 https://github.com/datajoint/datajoint-python/pull/1517

lochhh:
Change PR scope to set datajoint min version

lochhh:
Closing as requested change included in #587

PR #602, opened 2026-07-28
Fork-safe ephys parallelism and auto-delete of the preprocessed recording

esutlie:
## Summary
- Replace `n_jobs=-1` in the spike-sorting parallel steps with a cgroup-aware, capped worker count on a thread-based pool (`fork_safe_job_kwargs()`, which sets `pool_engine="thread"`), in `PreProcessing`, `PostProcessing`, and `SIExport`.
- Delete the preprocessed `recording.zarr` after a successful `SpikeSorting`, guarded so it only deletes when no sibling paramset for the same block and electrode group still needs it.
- Raise an actionable error if `SpikeSorting` runs when the preprocessed recording is missing (deleted after a prior sort), pointing at the re-populate recovery path.

### Why the parallelism change

Two independent problems, both triggered by `n_jobs=-1`, made the preprocessing step intermittently hang on the HPC:

1. **Oversubscription.** `n_jobs=-1` resolves against `os.cpu_count()`, which on SLURM returns every physical core on the node (64+), not the handful of CPUs the job was actually allocated (the cgroup limit). So a 4-CPU job would try to run 64 workers, thrashing the node. `fork_safe_job_kwargs()` instead reads the schedulable-CPU count (`os.sched_getaffinity`, i.e. the real cgroup allocation) and caps it, so we never spawn more workers than we have CPUs.

2. **Fork deadlock.** By default SpikeInterface parallelizes by *forking* the process — the child is a raw copy of the parent's memory. If the parent was mid-way through a NumPy/BLAS call at fork time, the child inherits a locked internal state it can never unlock, and the worker hangs forever. This is a known hazard of mixing `fork` with multithreaded native libraries, and it's intermittent because it depends on the exact moment the fork happens. Setting `pool_engine="thread"` makes SpikeInterface use threads instead of forking, which removes the failure mode entirely.

The same change applies to all three steps that used `n_jobs=-1`.

### Why delete the preprocessed recording

`recording.zarr` is a large, regenerable intermediate (~60% of the per-block footprint) that only the sorting step reads, so deleting it after a successful sort reclaims that space. Because it's shared by every paramset for a block and electrode group, deletion is skipped while any sibling paramset is preprocessed but not yet sorted; a not-yet-preprocessed sibling regenerates it through its own PreProcessing. To re-sort after deletion, delete the block's PreProcessing entry and re-populate.

## Validation

Validated on the SWC HPC against the golden ephys dataset on aeondj: the unit suite (143) and the ephys golden integration suite (33) both pass, including the PreProcessing step that previously hung.

### DIFF
--- a/aeon/dj_pipeline/spike_sorting.py
+++ b/aeon/dj_pipeline/spike_sorting.py
@@ -7,6 +7,7 @@ sorting, postprocessing, exports, and ingestion of sorted units/waveforms.
 import importlib
 import json
 import os
+import shutil
 import tempfile
 from collections.abc import Iterator
 from datetime import UTC, datetime
@@ -23,6 +24,7 @@ from aeon.dj_pipeline import get_schema_name
 from aeon.dj_pipeline.utils.ephys_utils import resolve_ephys_file
 from aeon.dj_pipeline.utils.paths import get_sorting_root_dir
 from aeon.dj_pipeline.utils.spike_sorting_utils import (
+    fork_safe_job_kwargs,
     resolve_analyzer_dir,
     strip_non_numeric_properties,
 )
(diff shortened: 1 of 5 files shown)


ttngu207:
LGTM!
The only risk is to truly confirm that `recording.zarr` is not used by any downstream steps except for the SpikeSorting. 
Looks like this is confirmed in our previous call. So we're good to proceed here.

ttngu207 (approved):

PR #604, opened 2026-08-20
Improve spike-sorting manual curation: additional tags, GUI-related improvements, cascade-delete fix, and zarr analyser support

zofiasus:
I propose the following changes, some are needed for the pipeline to work well, while others are helpful additions for easier manual curation:

## `spike_sorting.py`

**Using Unit Matching params**
Before, the matching params could be set in `UnitMatchingParamSet()` but weren't used anywhere. Now they're actually implemented in the matching, and if unset (empty dict) the previous defaults are used via `_DEFAULT_MATCHING_PARAMS` (delta_time=0.4ms, match_score=0.5, min_score=0.1).

**Saving the match_score matrix for more than just the matches**
New `UnitMatching.CandidateMatch()` part table — saves the match values for all candidates above `min_score=0.1`, for QC and maybe future investigation. `UnitMatching.Unit.matched_spike_count()` still only holds the winning match.

**Helper functions**
`_restrict_to_overlap`, `_load_block_unit_spike_trains`, `_compare_spike_trains_in_overlap` used to sit inside `UnitMatching.make()`, now live outside it — makes them testable, and lets the new params actually get used.

**Global unit key search fix**
When an unmatched unit needs a new global unit ID, the code checks the highest existing ID for this insertion and takes the next one. That check was filtering `GlobalUnit` by `insertion_key & paramset_key`, but `matching_paramset_id` is only a secondary attribute on `GlobalUnit`, not part of its primary key (`(ProbeInsertion, global_unit)`) — so filtering by it was wrong and could cause ID collisions across paramsets.

**Adding quality & tags to each unit**
`SortedSpikes.Unit.curation_quality` attribute (for 'noise/MUA/good' labels) and `SortedSpikes.UnitTag` part table (for 'irregular waveform', 'amplitude drift', etc. labels). Both get inserted in `spike_sorting_curation.py`'s `insert_sorted_spikes_from_analyzer()` — one `UnitTag` row per (unit, tag) pair, a unit with 3 tags gets 3 rows, none gets zero.

## `spike_sorting_curation.py`

**Defining tag options**
`MANUAL_TAG_OPTIONS` — read by `SortedSpikes.UnitTag`, and by `launch_si_gui.py`/`multitag_view.py` to make the tags available in the GUI and then downstream in the DJ tables.

**Fixing the deleted SortedSpikes cascade — in `ApplyOfficialCuration()`**
Applying a new curation means deleting the old `SortedSpikes` entry and inserting a new one. But `OfficialCuration()` inherits from `SortedSpikes()`, so deleting the parent deletes the child too :( Fix: save the full `OfficialCuration` row before the delete, let the cascade happen, recreate `SortedSpikes` from the curated analyzer, then re-insert `OfficialCuration` pointing at the fresh row before `ApplyOfficialCuration` inserts itself.

**Zarr format support — in `save_manual_curation()`, `ApplyOfficialCuration()`**
Previously only handled the binary-folder format (curation data in a plain `curation_data.json`). Zarr analyzers store it in the zarr group's `.zattrs["curation_data"]` instead, no JSON file. Now branches on `is_zarr` to read from the right place either way, but always writes the versioned snapshot as a real JSON file (since `ManualCuration.File` tracks an actual file on disk).

Bundled in here too:
- switched from copy-then-delete-original to write-fresh-snapshot-then-clear-original, since zarr's source isn't a file to copy at all
- DB inserts now use the fully-resolved `full_key` instead of the caller-supplied `key`, which was missing fields `ManualCuration`'s actual primary key needs
- reordered so DB inserts happen before clearing the live curation_data copy, so a failed insert doesn't leave an orphaned, unrecoverable snapshot
- `save_as(format="zarr")` silently appends `.zarr` to the folder name — the exists-check and path bookkeeping didn't account for that, fixed via `clean_zarr_folder_name()`

**File fetching fix in `ManualCuration.File`**
The curation JSON is `curation_data_id{curation_id}.json`, and `ApplyOfficialCuration` also writes a `"curation_applied_analyzer"` file under the same `curation_id` — so fetching by `curation_id` alone could match both. Lookup now also restricts on `file_name`.

**DJ 2.3.2 version fix — in `ApplyOfficialCuration()` & `restore_raw_sorting()`**
`.delete(safemode=False)` → `.delete(prompt=False)` — `safemode` doesn't exist on this DataJoint version anymore, would've hard-crashed.

**`launch_spikeinterface_gui()` — three new parameters**
`layout`, `label_definitions`, `with_traces` — what `launch_si_gui.py`'s `--layout`/`--no-traces` flags plumb into. Gives you: a custom layout from a JSON file, opening the GUI without raw traces (faster), and custom tags beyond noise/MUA/good.

**GUI-launch resilience — skip failed extensions instead of crashing**
Some data is incomplete — no PCA, no spike amplitudes computed, etc. This used to crash the whole GUI. Now it loads extensions one at a time, tracks the failures, and opens with whatever views are actually available :) Plus a smaller fix: `gui_dir.mkdir()` now gives a clear error instead of a bare traceback, so you know what's actually missing.

## `launch_si_gui.py`

**New CLI flags**: `--layout` (points at `spikeinterface_gui_layout.json`), `--no-traces` (drops raw trace views to reduce file-handle pressure on long blocks).

Example use: `uv run python submodules/aeon_mecha/aeon/dj_pipeline/scripts/launch_si_gui.py --layout submodules/aeon_mecha/aeon/dj_pipeline/scripts/spikeinterface_gui_layout.json --no-traces`

**Registers MultiTagView** from `spike_sorting_curation.py`'s `MANUAL_TAG_OPTIONS` into the GUI's view list, wires `label_definitions` (quality + tags) through to `launch_spikeinterface_gui()`. Optional feature, control if present by changing the layout.json file to include the 'multitag' view.

**Printing errors & info in the terminal**
Before, "[ERROR]: Uncaught exception" was all you got. Now there's more informative feedback like: failed save-to-analyzer (possible connection loss), failed GUI open due to file-open limit, and confirmation that curation actually saved.

**Small additions**: launch-time feedback (nice to know how long it took), a save confirmation that also gives a clear error if `/Volumes/aeon` drops mid-session, a window title patch so multiple open GUI windows are distinguishable in Cmd+Tab/Mission Control (very helpful), and an open-file-limit info print to advise `--no-traces`.

## New files

**`spikeinterface_gui_layout.json`** — a custom layout of the GUI, enables the multitag panel (not in the default).

**`multitag_view.py`** — makes the tags usable by adding checkboxes and keyboard shortcuts onto the unit list panel, easier for the curator to use. Also checks that its tag list matches the one defined in `spike_sorting_curation.py`.``


esutlie:
Looks great @zofiasus!

@ttngu207 @lochhh Please have a look when you have time.

ttngu207 (dismissed):

aeon/dj_pipeline/streams.py
lochhh, line 1: is this change intended?

tests/dj_pipeline/utils/test_xarray_codec_integration.py
lochhh, line 1: this file should no longer exist

tests/dj_pipeline/utils/test_xarray_codec_unit.py
lochhh, line 1: this file should no longer exist

aeon/schema/ephys.py
```diff
@@ -52,8 +52,21 @@ def read(self, file):
...
+            for _fmt in ("%Y-%m-%dT%H%M%SZ", "%Y-%m-%dT%H-%M-%S"):
+                try:
+                    epoch = datetime.strptime(chunk_info, _fmt)
+                    break
+                except ValueError:
+                    continue
+            else:
+                raise ValueError(
+                    f"Unrecognized HarpSync chunk timestamp: {chunk_info!r}"
+                )
```
lochhh: i think we can use [`swc.aeon.api.io.chunk_key`](https://github.com/SainsburyWellcomeCentre/aeon_api/blob/63fb47d661e11f978dedfe671c94c490100ce594/src/swc/aeon/io/api.py#L130-L148) for this?
```python
epoch = chunk_key(file)[1].tz_localize(None)
```

aeon/dj_pipeline/streams.py
```diff
@@ -69,3 +69,967 @@ class DeviceName(dj.Lookup):
...
+            + Chunk(s) started after Camera install time for Camera and not yet removed
+            """
+            return (
+                acquisition.Chunk.join(
+                    Camera.join(Camera.RemovalTime, left=True),
+                    semantic_check=False,
+                )
+                & "chunk_start >= camera_install_time"
+                & 'chunk_start < IFNULL(camera_removal_time,"2200-01-01")'
+                & 'device_name = "CameraTop"'
```
lochhh: This line was deliberately added in https://github.com/SainsburyWellcomeCentre/aeon_mecha/pull/604/changes/3d72cbe8c90a349597c92d631820e6faa9f2cb53 as a fix. But the real fix needs to live in streams_maker.py/load_metadata.py gen logic so one doesn't need to manually apply this fix on re-generating the streams.py module

lochhh:
ive also a "clean" version of this branch at https://github.com/SainsburyWellcomeCentre/aeon_mecha/pull/612 if this looks right, i can do a force push to this PR so the commit history is linear

aeon/dj_pipeline/scripts/multitag_view.py
```diff
@@ -0,0 +1,156 @@
...
+        for lbl in self.controller.curation_data["manual_labels"]:
+            if lbl["unit_id"] == unit_id:
+                return list(lbl.get("labels", {}).get(TAGS_CATEGORY, []))
+        return []
+
+    def _set_unit_features(self, unit_id, features):
+        manual_labels = self.controller.curation_data["manual_labels"]
+        for lbl in manual_labels:
+            if lbl["unit_id"] == unit_id:
+                lbl.setdefault("labels", {})[TAGS_CATEGORY] = features
```
ttngu207, line 101: The tagging mechanism itself is the right shape — one boolean property per option for a non-exclusive category is exactly what `SortedSpikes.make` reads back. But there's a bug in how `multitag_view` writes it that I'd want fixed before this merges: **any unit that gets both a quality label and a tag loses the quality label on save**, and `SortedSpikes` silently falls back to Kilosort's KSLabel.

**Why.** The GUI and the new view write into `curation_data["manual_labels"]` in two different shapes:

- `controller.set_label_to_unit` writes **flat** — `{"unit_id": 5, "quality": ["good"]}` (`controller.py:995`)
- `multitag_view._set_unit_features` writes **nested** — `lbl.setdefault("labels", {})[TAGS_CATEGORY] = features` (line 101 here)

So labelling a unit and then tagging it produces a mix:

```json
{"unit_id": 5, "quality": ["good"], "labels": {"tags": ["flag"]}}
```

On save, `CurationModel` validation branches on whether `labels` is present (`curation_model.py:124-130`):

```python
labels = manual_label.get("labels")
if labels is None:
    labels = set(manual_label.keys()) - {"unit_id"}   # flat -> picks up every category
else:
    manual_label["labels"] = {k: list(v) for k, v in labels.items()}   # nested -> only what's in here
for label in labels:                                   # iterates "tags" only
```

`labels` exists, so it rebuilds from the nested dict and never looks at the flat `quality` key. Checked against the pinned spikeinterface:

```
MIXED -> {'unit_id': 5, 'labels': {'tags': ['flag']}}                        quality gone
FLAT  -> {'unit_id': 5, 'labels': {'quality': ['good'], 'tags': ['flag']}}   both kept
```

Both orderings hit it — tag-then-label lands in the same mixed shape via the `ix is not None` branch.

**Fix.** Write in the same flat shape the controller uses, so the model normalises every category together:

```python
def _get_unit_features(self, unit_id):
    for lbl in self.controller.curation_data["manual_labels"]:
        if lbl["unit_id"] == unit_id:
            return list(lbl.get(TAGS_CATEGORY, []))
    return []

def _set_unit_features(self, unit_id, features):
    manual_labels = self.controller.curation_data["manual_labels"]
    for lbl in manual_labels:
        if lbl["unit_id"] == unit_id:
            lbl[TAGS_CATEGORY] = features
            return
    manual_labels.append({"unit_id": unit_id, TAGS_CATEGORY: features})
```

Worth adding a test that round-trips one unit carrying both a quality label and a tag — that's the case nothing covers right now.

esutlie:
Thanks @lochhh, super helpful! I pushed fixes for all of these and merged main (#611) in so the branch is current.

- **streams.py**: not intended, it was an accidental full regeneration. I reverted it to main's stub.
- **The `device_name = "CameraTop"` line**: that's from an earlier fix (3d72cbe) for cameras without position tracking, so it's unrelated to this PR and it's gone now with the streams.py revert. You're right that the real fix belongs in the `streams_maker`/`load_metadata` generation logic so it survives regeneration. That's a separate change from this PR.
- **test_xarray_codec_unit.py / test_xarray_codec_integration.py**: removed. The current `test_codec_*` files supersede them.
- **chunk_key**: done. I confirmed it resolves both the compact-UTC and dashed timestamp filenames identically to the old parser.

esutlie:
I pushed the fix for the tag/quality-label issue @ttngu207 flagged. `multitag_view` now writes tags flat (matching how the controller writes quality), so a unit with both a quality label and a tag keeps its quality on save. I added a unit test covering both the label-then-tag and tag-then-label orderings.

@ttngu207 @lochhh I think that covers everything from your reviews except the CameraTop point, which I've left out of scope (details below). Please take another look when you get the chance.

@lochhh @AdrianRoggenbach on the `CameraPosition` / `device_name = "CameraTop"` restriction, I looked into it and left it out of this PR. It's a pre-existing issue in the streams generator, not something this PR touches. That restriction only ever lived as a hand-edit to the auto-generated `streams.py`, so `streams_maker` drops it whenever the file is regenerated. Short term, if you regenerate `streams.py` you can just re-add that line by hand (as in 3d72cbe). The real fix is to have the generator or the catalog carry it so it survives regeneration, but that's a separate change we can pick up down the line.

I think we should squash-merge, so `main`'s history stays linear. @zofiasus once the re-reviews are in, could you give it one more test before you squash-merge into main?

PR #611, opened 2026-09-14
Fix ephys→HARP sync + spike/IMU gap drops (OpenEphys.Onix1>=0.4.0)

AdrianRoggenbach:
## Summary

This PR fixes five bugs in the ephys → HARP synchronisation:

- all ephys HARP times are 1 s late;
- spikes are shifted within each ephys chunk by the first spike occurrence time;
- spikes and IMU samples outside the HarpSync rows are silently dropped;
- ephys chunks whose first and last samples both fall outside the HarpSync rows are never ingested.

This PR reverts #592. Each fix is a local edit in the existing function, to keep the diff reviewable.

### 1. [Sync model fitted on a column that is 1 s late](https://github.com/SainsburyWellcomeCentre/aeon_mecha/commit/f860a4d34962322ea5f732625663a4812b4be02b) (reverts #592)

`HarpSyncModel.Reader` fits the ONIX clock → HARP regression on the CSV `Seconds` column (#592). Since OpenEphys.Onix1 0.4.0 ([open-ephys/bonsai-onix1#319](https://github.com/open-ephys/bonsai-onix1/pull/319)), `HarpSyncInputDataFrame.HarpTime` already adds the one second that the [HARP sync protocol](https://github.com/harp-tech/protocol/blob/main/SynchronizationClock.md) requires. The acquisition workflow adds it again when writing `Seconds`, so `Seconds == Value.HarpTime + 1` in every row.

The reader now fits on `Value.HarpTime`. `EphysEpoch.epoch_start`, `EphysSyncModel.sync_start/sync_end`, `EphysChunk.chunk_start/chunk_end`, `OnixImuChunk.synced_df` and `SyncedSpikes` all move 1 s earlier.

The reader also drops a final CSV row that has no line ending. When acquisition stops uncleanly the CSV writer's last 4 KB buffer is lost, and the final row can be cut mid-line. In one of our files that row still parses, but with a truncated `Value.HarpTime` that would corrupt the fit.

Evidence, independent of neural data: with the current `Seconds`-based model, headstage IMU (Bno055) yaw rate lags CameraTop SLEAP heading angular velocity by +1.00 s. The peak is at +0.98 to +1.02 s in 12/12 one-hour windows, with r up to 0.71 and r ≈ 0 at lag 0.

### 2. [`SyncedSpikes`: spike indices counted from the unit's first spike, not the chunk start](https://github.com/SainsburyWellcomeCentre/aeon_mecha/commit/8b0c684800f3f325bfb8bba76ce82f09fa1b5679)

`indices2syncedtimes` did `spk_ind -= spk_ind[0]`. Each unit's first spike in every ephys chunk was therefore placed on the chunk's first clock sample, and all its later spikes in that chunk moved earlier by the same amount. On one block (94 units, 10-minute chunks) the shift had a median of 1.9 s and a maximum of 593 s, and 28% of unit–chunk pairs were shifted by more than 10 s.

Indices are now offset by the chunk's start in the concatenated recording. Chunk membership uses `[start, end)` instead of `(start, end]`; the old version assigned a chunk's first sample to the previous chunk.

### 3. [`SyncedSpikes`: spikes outside `[onix_ts_start, onix_ts_end]` dropped](https://github.com/SainsburyWellcomeCentre/aeon_mecha/commit/b325ee09c6e8653b57e19fb8825df1d089cfefe6)

HarpSync CSVs are hourly with one row per second, so consecutive sync windows leave a 1 s gap. Spikes matched no model, and were silently dropped, if they fell:

- before an epoch's first row (1–2 s);
- in those gaps;
- after the last row (up to ~80 s after an unclean stop).

That was ~0.03% of spikes per block, and 0.29% on a block whose recording ran 64 s past its last row.

Each spike now uses the last of its chunk's sync windows that starts at or before it (the first window for earlier spikes), extrapolating the linear fit. In a held-out test, extrapolating 80 s beyond the fitted rows gave errors ≤ 0.15 ms. Windows come from the chunk's own `EphysChunk.SyncModel` links and so always from its epoch. This matters because the ONIX clock restarts each epoch and a block can span two epochs; the old block-wide window list only worked while those clock ranges did not overlap.

### 4. [`EphysChunk.ingest_chunks`: chunks not linked to the sync windows they need](https://github.com/SainsburyWellcomeCentre/aeon_mecha/commit/8d455a3eb601551f9ab63a53a9dd486cba8082c0)

A sync window was linked only if it contained the chunk's first or last ONIX sample, so a chunk with both samples outside all windows was skipped. Three of our amplifier files were never ingested for this reason: one 4.6 s chunk after an epoch's last row, and two 275 s recordings that start before and end after their only window.

The fix links every window from the last one starting at or before the first sample to the last one starting at or before the last sample. If the chunk ends before the epoch's first window, it links that first window.

### 5. [`OnixImuChunk.make`: IMU samples outside the sync window dropped](https://github.com/SainsburyWellcomeCentre/aeon_mecha/commit/83c494c0ebfeb3af544e96519a1a8a285a58720a)

This is the same gap problem as 3: each row kept only samples within `[onix_ts_start, onix_ts_end]`. Each sync window now owns the samples from its start up to the next window's start; the first window also owns earlier samples, and the last window later ones. `OnixStreamCodec` already filters on the bounds stored in `stream_df`, so it is unchanged.

## Tests

- `tests/dj_pipeline/test_ephys_ingestion.py`:
  - `test_sync_start_aligns_with_hour_boundary` now expects `XX:59:59`.
  - `test_sync_start_matches_seconds_column_not_value_harptime` is renamed to `test_sync_start_matches_value_harptime_column`, with the columns swapped.
  - `test_spike_times_within_sync_range` is renamed to `test_spike_times_within_chunk_range`, because spikes outside the sync rows are now kept.
- `tests/dj_pipeline/test_onix_imu_pipeline_integration.py` and `tests/dj_pipeline/utils/test_codec_integration.py`: the IMU sample-count assertions (`0 < n < 100`, "filter trimmed something") now check that no sample is lost (`sum == 200`, `== 100`).
- `tests/schema/test_ephys_reader_unit.py`: the fixture's Aeon timestamp is now `harp_time + 1`, as in real files, so the unit test pins the column.

## Notes

- **Data recorded with OpenEphys.Onix1 ≤ 0.3.0** (released before Oct 2024) has a raw, 1 s early `Value.HarpTime`. For that data neither this reader nor the one from #592 is guaranteed correct without knowing how that workflow wrote `Seconds`. Please check the ONIX package version of other ephys experiments ingested through `HarpSyncModel`.
- **Existing rows carry the 1 s offset** and need re-ingesting.

🤖 Co-created with Claude Opus 5


lochhh (dismissed):

ttngu207 (dismissed):
LGTM. Nice detective work — the IMU↔SLEAP lag check is a good validation! 

Two smaller comments:
1. Bugs 2, 3 and 4 have no test that fails without the fix — I reverted each one locally and the tests still stayed green. `test_spike_times_within_chunk_range` can't catch the index bug either way, since both old and new code keep spikes inside the chunk's own range. Spike-count conservation (same shape as your IMU check) plus a synthetic chunk-outside-all-windows case would do it.
2. Some guard for Onix1 ≤ 0.3.0 — `Seconds == HarpTime + 1` holds in both regimes, so the CSV alone can't tell us which column is right.

I can add enhancement to the tests in follow up PRs

lochhh:
thanks @ttngu207 ! [Add regression tests](https://github.com/SainsburyWellcomeCentre/aeon_mecha/pull/611/commits/b46e92999a9ce8892241cae4b00b7d6555535f83)

ttngu207 (approved):

ttngu207:
Reapproved again. @lochhh I think this is good to merge. Thank you!

